PR #11, opened 2016-02-12
A couple minor changes to sc.py

mnfienen:
Also check out the notebook for Freyberg Schur. Some cool things but also some stuff we should discuss. Check out the issue on pyemu_note....

Another thing to consider....when the dataframe is returned from `next_most_important_added_obs`, the base uncertainty value I'm unsure about. Seems like it could:
  1.  Always be the same as the initial based before iterations
  2.  or it could change with each successive iteration such that the `next` observation evaluated is with respect to the previous iteration.

Not sure how that was reported up to now, but something to think about.


mnfienen:
Incorporated new `next_most_important_added_obs` edits from Jeremy and also removed transport predictions from freyberg model

PR #26, opened 2017-09-19
Add basic Sphinx Makefile and *.rst files

smwesten-usgs:
These commits added basic (as in bare minimum and barely working) files needed to build documents using Sphinx. 'conda install sphinx' should be sufficient to get it working with a typical Anaconda Python distribution.

Need to run 'make html' from within the 'docs' subdir to update docs from docstrings. 'conf.py' has been modified from the default values to include 'sphinx.ext.napoleon', which gives Sphinx the ability to parse Numpy and Google-style docstrings. 'Author' and 'version', 'release' numbers should probably be changed in 'conf.py', as these were just placeholder values for use when trying to get Sphinx working.

The resulting auto-generated document will be somewhat ugly until docstrings are consistently converted over to Numpy or Google-style tags.


jtwhite79:
Steve - you da man!  Thanks for this...now I have no excuse not to do a better job with doc strings

mnfienen-usgs:
uh -oh ---- docstring sprint in our future?

jtwhite79:
I don't think "sprint" appropriately represents how much work this is going to be...more like 100-mile ultra marathon.

PR #35, opened 2018-04-24
Develop

hwreeves-USGS:
Added a base case entry to the dataframe build by the jac_test helper static method


jtwhite79:
Sweet as, Howard.  Thanks man

PR #38, opened 2018-05-07
fixed generation of random choice 

hwreeves-USGS:
Makes an array of observations to use for plotting jacobian test results given a 
maximum number of output pages for each parameter.


jtwhite79:
thanks Howard!

Issue #40, opened 2018-05-12
dictionary key mismatch in fac2real?


fluidmotion:
probably abusing this - but i don't yet see how the dictionary key for pp_dict would match the key for fac_data?

- looking around line 1986 in geostats.py
`fac_sum = sum([pp_dict[pp] * fac_data[pp] for pp in fac_data])`

the pp_dict key is defined as the pilot point name
`pp_dict = {name:val for name,val in zip(pp_data.index,pp_data.parval1)}`

but the fac_data key is defined by some value preceding the factors? so i keep getting a 'key error'.
perhaps it's my inputs (i'm providing fac2real a dataframe rather than a pilot point file) - but it seems to reading it ok, and the required columns are present.

jtwhite79:
The factor lines in the factors file are pairs of integers and floats, where the integer in the index of the pp name listed at the top of the file and the float is the kriging factor.  This is how the pest utils do it too.  So, like the pest utils, it assumes the pilot points are listed in the same order in the pp file and the factors file  - prob should deal with the case where pp file is not in the same order as the factors file.  But for your case, I'm not sure what is up - in the factors file, are the same number of pilot points listed at the top compared to the pilot points dataframe?  If the pp dataframe and factors file have the same pilot points in the same order, you should be able to set an integer index (zero-based) on the dataframe.  If that doesn't work, can you upload your factors file and the pp dataframe you trying to use?

fluidmotion:
closed pull #41 related to this issue.
 but before we go... i'm still seeing some oddness after resetting index the dataframe (so that the index value matches the index of the factors). when this df is passed to the OrdinaryKrige constructor(?) - that df gets assigned to self.point_data, but in doing so, it appears to lose the reset index. i tested using
`self.point_data = point_data.copy()`
and it seems to work, but i haven't ran the autotest stuff yet.

jtwhite79:
There's a pretty high probability that the index of the ``OrdinaryKriege.point_data`` is getting reset somewhere, so it's good idea to copy ``point_data`` in the constructor. Did the tests pass?

fluidmotion:
addressed in c2bc47fd243ac0b887d935a823cbb96989ce8e13 - closing.

PR #41, opened 2018-05-12
potential fix for key mismatch in issue #40

fluidmotion:
- relating pilot point names to factors in geostats.py

### DIFF
--- a/pyemu/utils/geostats.py
+++ b/pyemu/utils/geostats.py
@@ -1983,9 +1983,9 @@ def fac2real(pp_file=None,factors_file="factors.dat",out_file="test.ref",
             raise Exception("error parsing factor line {0}:{1}".format(line,str(e)))
         #fac_prods = [pp_data.loc[pp,"value"]*fac_data[pp] for pp in fac_data]
         if itrans == 0:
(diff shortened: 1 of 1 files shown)


jtwhite79:
This PR fails all of the fac2real related tests, including the verification tests against the pest uitls = frowny face...Let's see if we can get your case going, then we can revisit refactoring the fac2real function to use the actual names of the pilot points.  I'm just worried about the additional level of indirection: factor index -> pp index -> pp name -> pp value; this function is already slow compared to the compiled version...

fluidmotion:
ugh. just saw that. oops. i believe the pilots and the factors are listed in the same order, but the factors appear to begin with 1? whereas the pilots use a feature ID from a subset of a GIS file (starting at 17334). i suppose the subset can be reset to one-based, just didn't realize the need.

jtwhite79:
The factors are 1-based but internally, pyemu converts that zero-based (the real base!), so you prob want to have a zero-based index on the subset...

fluidmotion:
ah - i didn't see this 
`pp_dict = {name:val for name,val in zip(pp_data.index,pp_data.parval1)}` 
where the pp_dict key is the index, rather than the 'name' column.
sorry for the trouble. closing.

PR #53, opened 2018-07-06
Adding support for defining desired sfr pars to setup in PstFromFlopyModel

briochh:
Can now pass a list of desired sfr pars to the sfr_pars argument in helper PstFromFlopyModel().

Usefull if you don't need all of the pars to be setup. 

Updating testing to test for various scenarios of lists being passed. Raising many warnings if pars are asked for that dont exist in the sfr.


jtwhite79:
Thanks B - Appveyor is failing intermittently because the max time for open source (free) CI is 1 hr and we right around that time limit currently.   Lets think about if we want a subset of tests to run on Appveyor (prob get away with just the full meal deal tests for now...)

briochh:
erggg, thats frustraiting. Is there a lot more going on with appveyor than travis for it to take twice as long to complete? Subsetting tests makes me a little nervous!

jtwhite79:
Not sure why the tests are so much slower on appveyor...makes me nervous too, esp give  the recent os discrepancies we saw with pandas

langevin-usgs:
Pandas discrepancies?  Hmm.  That's strange.

PR #54, opened 2018-07-20
Adding support for passing 'skip' as array to gw_utils.setup_hds_obs()

briochh:
Also added check to prevent obs being setup for indicies (kper,k,i,j etc) in apply_mtlist_budget_obs(), if they are returned by flopy.utils.MtListBudget of MfListBudget (they might not me at the moment).
A few aditional tests for these changes.
Also added a 1 second sleep in the helper, after removing exisintg directory. Seems to help with errors relating to windows holding on to the deleted directory and preventing writing to it.


jtwhite79:
Thanks B!

PR #55, opened 2018-07-26
Mod to sfr par setup, prevent duplicate apply

briochh:
Prevent duplicating sfr par apply in froward run where reachinput pars
are specified (my bad). Now sets a bool. flag and calls
frun_pre_lines.append just once, writing the more flexible
`pyemu.gw_utils.apply_sfr_parameters(reach_pars=bool.)` to the frun
script.


briochh:
@jtwhite79 think this one is good to go now.

PR #62, opened 2018-08-24
support defining different k_zone_dict for different paramters by passing a dictionary of k_zone_dict

briochh:
A little nervous about this one so might be best to incorperate as feature branch until we have had a chance to properly break it. 

I wanted to add support for using different k_zone_dicts for different pars. the approach that I have taken is to allow passing `{pak.attr: k_zone_dict, pak2.attr: k_zone_dict2}`. Worked a beauty for my problem but then realised I needed to chase it through the pilot point prep and setup methods. This is where my nerves have kicked in. I have added a couble of additional test scenarios - but not certain that I am catching everything. 

One big potential issue I see in the future is cross-referencing which pak.attr k_zone_dict (e.g. `id='hk'`) belong to which pargp (e.g. `id = 'hk0'`) currently using a `str.startswith()` but may need to look into a more robust method.


briochh:
I have been running my `feature` branch for a little while now and it is working well - it's really handy to be able to define different zones for different paramters types. I have recently started using it with pilot-point parameterisation which is where I was concerned that there might be issues but that also appears to be working ok. 

Anyway, if you (and Travis) are happy with it it should be safe to merge.

jtwhite79:
I'm always happy to accept PR's, but Travis is picky!  The test that is failing is a result of this flopy change: [https://github.com/modflowpy/flopy/pull/420](https://github.com/modflowpy/flopy/pull/420)
I've patched it in the pyemu `data_assimilation` branch - its a really simple fix, just need to change 
``load_only=[]`` to ``load_only=["BAS6"]`` on line 1080 in utils_tests.py.  Can you make that change and resubmit PR?

briochh:
Oh yeah! That one keeps catching me out!

jtwhite79:
Hey @briochh - Haven't forgotten about your PR - just trying to get the issue with the pyshp 2.0 API change sorted out first.

jtwhite79:
Thanks @briochh!

PR #65, opened 2018-11-16
Correcting some messyness with specifying sfr_pars in ph setup

briochh:
Previous behaviour was trying to setup segment pars even if only
paramters in reach_data were requested in
`pyemu.helpers.PstFromFlopyModel(sfr_pars=["pars"])`. Also the par setup
wasn't finalising if only reach_data pars were setup, the
`apply_sfr_pars()` line was being ommited from frun file. I think the blame soley rests with me for that one!


jtwhite79:
Thanks B!

PR #69, opened 2019-02-26
Marrying temporal SFR pars with reachinput options and specifying sfr_pars options.

briochh:
new temporal sfr pars shouldn't effect setup of reach pars as reach pars
are just in dataset 2 for SFR, tempoal pars should apply to only segdata
(data set 6/4).

So just a minor change to support case where a reach parameter is listed
in sfr_pars (passed to `PstFromFlopyModel()` as `sfr_pars` or
`setup_sfr_pars()` as `par_cols`)

Also added test in `from_flopy_reachinput()` to add
`temporal_sfr_pars=True`.

### DIFF
--- a/autotest/pst_tests.py
+++ b/autotest/pst_tests.py
@@ -688,10 +688,10 @@ def from_flopy_reachinput():
         return
     import pyemu
 
-    if platform.platform().lower().startswith('win'):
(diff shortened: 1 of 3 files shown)


jtwhite79:
Thanks B!  something weird going on with the (old) intel linux pestpp binaries - sorting now...

PR #79, opened 2019-06-07
Speedup for sfr obs setup

briochh:
I've made some changes to the `load_sfr_out()` method to try and speed up the setup of sfr obs.

I think we can get away without `.loc` for these assignments, which speeds things up a bit. The biggest gain comes from using a list comprehension instead of a `df.apply(lambda)` fuction to construct the segment reach strings. 

These little changes seem to make a reasonable difference, as they get hit a lot of times when setting up sfr obs with many output times (the sfr output is read twice and these lines are hit for every output time block). Setting up the sfr obs now seems to be around 30% faster but I am sure there are more savings that can be made with similar changes elsewhere.


jwhite-usgs:
thanks B!

PR #86, opened 2019-07-28
histogram plots require setting xlim or figure sizes are wonky

wkitlasten:
should resolve issue #83 regarding funky figure sizes


jwhite-usgs:
thanks @wkitlasten !  much appreciated bru

PR #92, opened 2019-09-16
switching up to long names for mt list observations

briochh:
following on from mflist obs > 20.


jtwhite79:
Thanks for this @briochh .  Did you maybe forget to “git add” that new mt3d list file?

briochh:
yeah, good catch @jtwhite79 . :facepalm:

PR #94, opened 2019-10-08
add divide to ensemble

briochh:
Maybe this was omitted from the ensemble refactor for a reason but trying to add in supporting divide function on ensemble. 

Would it be worth adding a `get_dataframe()` wrapper to the ensemble class (or a de-protected `.df`)? I might be missing something but I think that `_.df` is the only way to completely pull the dataframe object out of the ensemble class object...

thoughts?

### DIFF
--- a/pyemu/en.py
+++ b/pyemu/en.py
@@ -102,6 +102,12 @@ class Ensemble(object):
         except:
             return self._df * other._df
 
+    def __truediv__(self, other):
+        try:
+            return self._df / other
+        except:
+            return self._df / other._df
+
     def __add__(self,other):
         try:
             return self._df + other


jtwhite79:
Thanks @briochh .  Just an oversight on divide.  Id like to keep the data frame as semi-protected/private just to encouarage use of the ensemble methods (and I think the getter and setter overrides should help make most things seamless).  But you can always access ._df - is that ok?

briochh:
Yup, sounds good @jtwhite79 . I've just been running into some issues with my old scripts, processing output ObsevationEnsembles where I only really need the dataframe. I have been caught out using lazy/basic dataframe indexing (straight `[]`) rather than `.loc[]` (I guess this could be supported with an upper level `__get_item__()` override but not sure if this would produced unintended consequences).

Anyway `._df` will work fine for me, as long as it doesn't offend you too much!

PR #97, opened 2019-10-24
Chunk files to speed up apply_array_pars() parallel processing

briochh:
Added a little wrapper to breakdown and process `model_file`s in chunks in `apply_array_pars()`.

I think this gets greater benefit from the parallelising of the processing in `apply_array_pars()` method. Instead of spawning a processor (or thread - unsure of the correct parlance!) for every model input file,  we just break the list of files down into chunks (of 50 at the moment) and let each spawned process work on it's chunk. -- Less time spawning more time calculating (I hope). 

The effect isn't too noticeable in the test case in `pst_tests_2.py` but I have trialled it on a local problem with 840 unique model files: `apply_array_pars()` was taking over 30 minutes; now less than a minute. The efficiency saving may not be so high where more `mults` are applied per `model_file` but generally, I think it should be faster.

### DIFF
--- a/autotest/pst_tests_2.py
+++ b/autotest/pst_tests_2.py
@@ -660,7 +660,7 @@ def at_bounds_test():
 
 if __name__ == "__main__":
 
-    at_bounds_test()
+    # at_bounds_test()
     #process_output_files_test()
     #change_limit_test()
     #new_format_test()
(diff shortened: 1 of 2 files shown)


jwhite-usgs:
Thanks B.  The chunking makes things marginally slower for a case with lots of multiplication but if its that much faster for your case, then its a net benefit!

briochh:
Net gains. If you can send me a good example case with lots of mults, I can take a look at improving at some point. I think a bit of preconditioning how those lists are divided might help.

PR #98, opened 2019-10-25
 Fixing issue with `pst_path` in `pst_from_io_files()`

briochh:
The new support for specifying `pst_path` breaks pest/pyemu where apply_array_pars() (and similar) are expecting mult files to be in subdirectories (which is the result of the `_setup_..._pars()` functions), as it strips the file basename from the path. As a result the input files that PEST generates are created in the wrong directory and are then not available for the `apply_array_pars()` functions. 

I guess this slipped through the tests because there are still (dummy) mult files in the mult directories - but these wont get updated by PEST!

This PR seems to fix this issue but I am not totally across the aim/logic for the addition of the `pst_path` option so there is a chance that I might have broken how this works. - not sure what the "catch all" option is for this.


jwhite-usgs:
wow - thats a big bust.  Nice catch @briochh

PR #129, opened 2020-05-14
zero basing error with new fmt binary jco causing row offset

MJKnowling:
Need to double check issue not at the `PESTPP-GLM` and `PESTPP-OPT` end

Let's see which tests are touched..

Can add test later too if needed


MJKnowling:
Confirmed at pestpp end. Fixed in pestpp release 4.3.*

PR #138, opened 2020-06-09
Fixing add_obs issues +

briochh:
Was seeing an issue with instruction file mis-alignment with desired observations - so attempting a fix here together with some tests to check that the values picked up are as expected. The instructions file builder is a bit messy, so may be worth a check @jtwhite79.

Also fixed `use_rows` option in `add_observations()`.

Also added quick fix to @mnfienen-usgs spatial ref check. Mike, can you check that this is catching what you want?


jwhite-usgs:
Nice one B!

mnfienen-usgs:
Looks good to me @briochh! I've had trouble reproducing the error to test it, but the code def looks on point (and caught my failure to loop over all the relevant files). Thanks!

PR #139, opened 2020-06-09
Feature pstfrom dev

jwhite-usgs:
bug fix in new mp.Pool usage


briochh:
@jtwhite79,  good call with the `chunk_len` argument. The savings gain through chunking the multiprocessing stuff was generally platform dependent with the old `mp.Process`. I believe on windows the whole environment needs to be bundled and distributed when spawning processors == slow spawning. Is there still this difference in behaviour when using `pool.apply_async`?

jtwhite79:
I only added the chunk arg so I could test what happens with chunk = 1!   But would be nice to thread that on thru to the PstFrom level.  I’m guessing pools performance is similar to process?

PR #142, opened 2020-06-13
Feature bounds_report

mnfienen-usgs:
New method that makes a dataframe summarizing parameters at bounds, by group and total for iterations.

### DIFF
--- a/pyemu/pst/pst_handler.py
+++ b/pyemu/pst/pst_handler.py
@@ -1,6 +1,7 @@
 
 from __future__ import print_function, division
 import os
+import glob
 import re
 import copy
 import warnings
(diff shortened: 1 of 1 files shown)


mnfienen-usgs:
oops. Hey @jtwhite79  are the par files still <pstroot>.par0, <pstroot>.par1, etc. for GLM? I was assuming a form of <pstroot>.0.par, <pestroot>.1.par, etc. which is like iES. If the former, I'll need to adjust this method to handle that

jtwhite79:
Honestly, I’m inclined to rename the glm par iteration files to be more consistent with ies (and easier to parse) so let’s keep this as is...

mnfienen-usgs:
sounds good. plus, I like having a consistent extension. easier to `glob` for

Issue #140, opened 2020-06-10
`PstFrom` support for multipliers AND base par vals?


mnfienen-usgs:
So, @jtwhite79 and I were discussing the potential need to allow support for base parameters in addition to multipliers when using `PstFrom`. For sure, both have good use.

One thing that would be ideal, in my opinion, would be to allow users to still parameterize some property in multiple ways (e.g. const, zone, pp, grid) but assign one of those to be at a base value and the others to be mults. That could get after things like estimating the mean of a field (say base is a const) but then add some variability about that mean (with grid or pp as mults). 

This would take some shenanigans of file handling since only one array/list would be put in org and others would be written as base or mults and (optionally) multiplication would take place. Maybe a holding /tmp folder? Not sure how best to handle. I like the idea, though, of fixing all the mults and casting a problem in real base values and then activating the mults as appropriate.

Anyway, just staring a convo here. @briochh and @MJKnowling  - what do you think?

mnfienen-usgs:
haha! you guys give me too much credit. In any case, a couple thoughts

0. I agree we should probably not stomp on the files in `org` so I think in cases where direct pars are used, we maybe rely on the TPL file to write the "base" file in org (but don't actually keep the original model files) and then apply mults as we normally would

1. As for having multiple methods to add direct vs. mult pars, makes sense, but I also think it will be important to enforce a single direct par set for a given list or array. I guess we could keep track in either case in the parent object.

jwhite-usgs:
@briochh - Im getting started on this today.  I'll submit a PR to the `feature_pstfrom_dev` branch once I have something that seems like it might kinda sometimes maybe work as expected.

mnfienen-usgs:
looks like a good start - thanks for jumping on it!

PR #144, opened 2020-06-16
Feat direct pars

jwhite-usgs:
added `par_style` optional arg to `PstFrom.add_parameters()` that can either be "multiplier" (default) or "direct" - addresses #140 .  Direct pars are setup to write the original model input file in the `org` dir using template files, so standard PEST interface stuff.  Currently supports constant, zone, and grid array-based pars and constant and grid list-based pars.  Too lazy to do the others, but the ground work is there.  A new mf6 test was added that verifies that direct recharge and wel pars are being written as expected for a single edge case.  Also patched an issue with calling `build_pst()` repeatedly where the model command is getting added repeatedly


briochh:
Looking good

PR #147, opened 2020-06-17
Feature pstfrom dev

jwhite-usgs:
added `Pst.try_parse_name_metadata()` to automatically parse the `PstFrom` par and obs name scheme into dataframe columns


pyemu/pst/pst_handler.py
```diff
@@ -3005,3 +2804,29 @@ def get_adj_pars_at_bounds(self, frac_tol=0.01):
...
+
+        """
+        par = self.parameter_data
+        obs = self.observation_data
+        par_cols = pst_utils.pst_config["par_fieldnames"]
+        obs_cols = pst_utils.pst_config["obs_fieldnames"]
+
+        for df,name,fieldnames in zip([par,obs],["parnme","obsnme"],[par_cols,obs_cols]):
+            items = df.loc[:,name].apply(lambda x: [item.split(':') for item in x.split('_') if ':' in item])
+            meta_dict = items.apply(lambda x: {k:v for xx in x for k,v in x})
```
briochh, line 2822: Some of these apply functions can get suuuuuper slow when the DataFrames are long. I think that these can be condensed into one apply.
```suggestion
            meta_dict = df.loc[:,name].apply(lambda x: dict([item.split(':') for item in x.split('_') if ':' in item]))
```
?

jwhite-usgs: nice!  I had them combined originally, but, in the course of debugging my hackery, I undid it - sad

briochh: nailed it

PR #154, opened 2020-07-09
feat(helpers): added quantile selection for observation ensembles

mnfienen-usgs:
Added a function in `utils.helpers.py` to return a quantile realization from an ensemble. It's calculated by finding the qth quantile as requested at each point in the ensemble (this is not an actual realization, though - too smooth) and then finding the actual realization with the smallest distance in a weighted sum of squares sense from that vector. 

Returns an augmented ensemble dataframe with the quantile realizations repeated at the end labeled with their quantile value. Also returns a dictionary with `real_name` for each quantile to be used if user wants to access the parameter sets corresponding to those quantiles in the observations.

User can optionally provide either a list of observation names  or groups to be used in the evaluation (note - in any case, the entire ensemble is returned, but quantile calculations are only made using the subset if it is selected). This allows a user to potentially focus in on a quantity of interest, or a subset of interest.

### DIFF
--- a/pyemu/utils/helpers.py
+++ b/pyemu/utils/helpers.py
@@ -310,6 +310,19 @@ def geostatistical_prior_builder(pst, struct_dict, sigma_range=4,
                 #     print(full_cov.names[i])
     return full_cov
 
+def _rmse(v1, v2):
+    """return root mean squared error between v1 and v2
+
+    Args:
+        v1 (iterable): one vector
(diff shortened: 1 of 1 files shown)


mnfienen-usgs:
Added another helper to calculate unweighted RMSE between all ensemble members and observed values. Can calculate by group or only total. Can also return for only a subset of realizations.

PR #166, opened 2020-08-19
Feat pstfrom name cast

jwhite-usgs:
Casting user supplied base names and group names to lower. Added default ult bounds.

### DIFF
--- a/pyemu/utils/pst_from.py
+++ b/pyemu/utils/pst_from.py
@@ -851,6 +851,7 @@ class PstFrom(object):
                 use_rows = df.iloc[use_rows].idx_str.unique()
             # construct ins_file from df
             ncol = len(use_cols)
+
             obsgp = _check_var_len(obsgp, ncol, fill=True)
             df_ins = pyemu.pst_utils.csv_to_ins_file(
                 df.set_index('idx_str'),
@@ -1159,6 +1160,12 @@ class PstFrom(object):
                                "single-element container, or container of "
                                "len use_cols, not '{0}'"
                                "".format(str(par_name_base)))
+
+        # otherewise, things get tripped up in the ensemble/cov stuff
+        if pargp is not None:
+            pargp = pargp.lower()
+        par_name_base = [pnb.lower() for pnb in par_name_base]
+
         if self.longnames:  # allow par names to be long... fine for pestpp
             fmt = "_{0}".format(alt_inst_str) + ":{0}"
             chk_prefix = "_{0}".format(alt_inst_str)  # add `instance` identifier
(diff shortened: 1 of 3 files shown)


pyemu/utils/pst_from.py
```diff
@@ -150,25 +152,10 @@ def parfile_relations(self):
...
-                                       pd.Series([None]) if x.isna().all()
-                                       else x[x.notna()].values).T
-                    self.logger.warn("Upper bound for par passed for some but "
-                                     "not all instances, will set NA to "
-                                     "passed values\n{}".format(fil))
-                    # replace Nones in list in Series with passed values
-                    pr.loc[g.index, 'upper_bound'] = g.use_cols.apply(
-                        lambda x: [fil[0].loc['ubound{0}'.format(c)] for c in x]
-                        if x is not None else fil[0].loc['ubound'])
+                ub_min = ubound.min().fillna(self.ult_ubound_fill).to_dict()
```
briochh, line 155: Wonder if min() and fillna() need to be the other way around? Can't remember if pandas returns na as the minimum if the series contains na, and we want to make sure the 'None's are set to a value if one is ever passed. NB. Are we happy skipping through with the min passed value if the user specifies more than one bound for a given par?

pyemu/utils/pst_from.py
```diff
@@ -178,20 +165,10 @@ def parfile_relations(self):
...
-                    fil = lbound.apply(lambda x:
-                                       pd.Series([None]) if x.isna().all()
-                                       else x[x.notna()].values).T
-                    self.logger.warn("Lower bound for par passed for some but "
-                                     "not all instances, will set NA to "
-                                     "passed values\n{}".format(fil))
-                    pr.loc[g.index, 'lower_bound'] = g.use_cols.apply(
-                        lambda x: [fil[0].loc['lbound{0}'.format(c)] for c in x]
-                        if x is not None else fil[0].loc['lbound'])
+                lb_max = lbound.min().fillna(self.ult_lbound_fill).to_dict()
```
briochh, line 168: Ditto above, and should that one be 'max()' for lbound?

pyemu/utils/pst_from.py
jwhite-usgs: I thought about this too. I  settled on reduction then fill because that means we sort out which user-supplied value to use, then fill any of the remaining all-nan entries with the fill value.  Im not sure which of min/max to use for upper/lower so I just picked min-upper and max-lower so that we dont choose extreme values over more meaningful values.

pyemu/utils/pst_from.py
jwhite-usgs: correct

pyemu/utils/pst_from.py
briochh: Yeah, that works, looks like pandas ignores NaNs for min/max operations (differs from numpy here). min-upper, max-lower makes sense to me.

PR #189, opened 2020-11-12
Feat pst_from: Flexible pathing for input array and list files; option to write tpl files to a subfolder

aleaf:
This pull request is intended to allow users to input files to ``PstFrom.add_parameters`` that aren't at the root level of the model workspace. It also includes an option to write template files to a subfolder (via a ``tpl_subfolder`` argument to ``PstFrom.__init__()``).

This is mostly implemented by refactoring pathing to use ``pathlib`` instead of ``os.path``. The following types of pathing situations should work (see the tests in ``autotest/pst_from_tests.py:TestPstFrom`` for more details):

```python
            Path('hk0.dat'),  # sim_ws; just file name as Path instance
            'hk1.dat',  # sim_ws; just file name as string
            Path(self.sim_ws, 'hk2.dat'),  # sim_ws; full path as Path instance
            'external/hk3.dat',  # subfolder; relative file path as string
            Path('external/hk4.dat'),  # subfolder; relative path as Path instance
            '../external_files/hk5.dat',  # subfolder up one level
```
@jtwhite79 @briochh Let me know what you think!


jtwhite79:
Thanks @aleaf !  I'm gonna kick tires a little tomorrow on a few active use cases but this looks excellent!

PR #194, opened 2020-11-25
Dirty workaround for some string actions on pathlib obj.

briochh:
Quickfix for issue with version 2 write of pest control file in PstFrom `build_pst()`.

The issue relates to a few `str` actions that are undertaken with the model path now failing when that path is a `pathlib.Path` object. There is probs a more designed solution for this but I am just calling os.path.join() over the `Path` object to convert back to string. Unsure if this will cause issues, perhaps @aleaf has some thoughts on this?


aleaf:
Hey @briochh- any `Path` instance can be converted to a string by wrapping it with the `str()` built-in (e.g. `str(new_filename)`. I actually only started using pathlib a few months ago and was a little skeptical as to its value at first, but am now convinced it is the way to go. Not that everything can't be done with ``os.path`` (and some cumbersome platform wrangling when needed), but pathlib just works, usually with more minimal, readable code. 

Anyways, in figuring it out, I found the [pathlib doc](https://docs.python.org/3/library/pathlib.html) to be pretty helpful. For the most part, ``Path`` instances are treated the same as string-based paths in python. It would be nice to implement this in pyemu as well. I'd be happy to help with this, either ad-hoc through pull request as things come up, or more systematically if we can come up with an inventory of all the places that would need refactoring.

PR #200, opened 2020-12-03
Feature ghactions

briochh:
Adding functionality for CI with GitHub Actions.

Currently running tests for linux on py 3.6, 3.7, 3.8 and Mac and Windows py3.8


briochh:
@jtwhite79 can you check out the travis fails in your time zone. If they just related to executable availability the. I think we can tolerate and merge.

jtwhite79:
it does seem to be a path thing - test passes locally for me

PR #211, opened 2021-01-11
Trying to speed up insfile processing and setup 

briochh:
A few attempted improvements to lines that are getting hit hard when instruction/output files have heaps of lines. 

Incl: 
using `x == y` rather than `x.startswith()`, 
dropping some superfluous pandas apply functions,
even using f strings rather than .format()!

Looking like saving 30-50% on insfile processing time.


jtwhite79 (approved):
looks good B!  nice one

PR #214, opened 2021-01-19
Feat chunks

jtwhite79:
wove the `chunk_len` arg into the `PstFrom` constructor and also added logic to not spawn an mp.Pool if there is only one chunk.  So to disable multiprocessing, just pass a really large `chunk_len`


briochh (approved):
tidy J

PR #215, opened 2021-01-22
Hotfix noheader list

jtwhite79:
added support for whitespace delim list files with and without headers.  Switched from `w` instruction to `!dum!` instruction in `pst_utils.csv_to_ins_file()` for whitespace delim since it should be a safer option to guard against leading whitespace issues.  Also relaxed the restriction that the `!dum!` entry in the output file be cast to double (this has also been changed in pest++ and I think pest will also make the change soon) - this helps with tables of mixed types that might not be numeric.  Had to switch to the `feat_newtools` branch of pest++ to support this so we need to keep in mind to switch back to `develop` at some point in the future.  One thing Im still wondering about: for rows that are to be skipped in the output file, should we skip the row with just a `l1` rather than filling out the row for all the markers and/or `!dum!` instructions?  I think so but @briochh would know better.


autotest/pst_from_tests.py
```diff
@@ -305,7 +304,8 @@ def freyberg_prior_build_test():
                  longnames=True, spatial_reference=sr,
                  zero_based=False)
     pf.extra_py_imports.append('flopy')
-    pf.mod_sys_cmds.append("which python")
+    if "linux" in platform.platform().lower():
+        pf.mod_sys_cmds.append("which python")
```
briochh, line 308: Think I just added this when I was having some pathing issues can probs be dropped if not useful.

briochh (approved):

briochh:
@jtwhite79, Re rows that are not to be observationalized (love it.): I agree that skipping whole rows is best. I think that this speeds up the processing of those HUGE files when not all rows contain obs. Is this what is currently happening (i.e we write `l1 \n`)? Could always use `l2` etc but probs getting unecessarily complicated.

jtwhite79:
Currently, each skipped row is getting a full compliment of 

```
l1 ~,~ ~,~ ~,~ ~,~ and on and on and on....
```
so skipping those rows with just an `l1` should speed up pyemu and pest (and make things more robust against junk on those skipped rows)

briochh:
double agree

PR #223, opened 2021-03-16
Feat inputstats

jtwhite79:
added function to helpers to calculate summary stats of array-based model inputs. works with PstFrom's mult2model_info.csv, which required sneaking in zone arrays (and writing out zone arrays) from PstFrom.add_parameters(). added a simple usage case to pst_from_tests.mf6_freyberg_varying_idomain().  summary stats include mean, min, max, std, median, at upper and lower bounds, and quantiles.


briochh:
Looks good J, shouldn't break anything anyway. Let's see how usage goes.

PR #228, opened 2021-04-05
added gsf reader to helpers

rosskush:
Loads a standard GSF file for extracting geospatial data for USG models


jtwhite79:
Thanks @rosskush !

PR #229, opened 2021-04-07
Feat pstfrom usg

jtwhite79:
added some basic support for using `PstFrom` with modflow-usg.  Its pretty hackey places, esp for using array-based parameters since usg still have the layer concept but doesnt include the node number for array-based inputs. So there is some trickery in the test function where you have to build separate sr look up dicts for each layer but the node number has to start at zero for each layer instead of the actual node number.  Added support for pilot points but only thru predefined locations.  Also added support for the `gw_utils.setup_hds_obs` now that flopy has a HeadUFile class.   Im open to suggested improvments!


pyemu/utils/geostats.py
```diff
@@ -1446,34 +1446,41 @@ def _worker(
     def to_grid_factors_file(
         self, filename, points_file="points.junk", zone_file="zone.junk"
     ):
-        """write a grid-based PEST-style factors file.  This file can be used with
-        the fac2real() method to write an interpolated structured array
+        """write a PEST-style factors file.  This file can be used with
+        the fac2real() method to write an interpolated structured or unstructured array
```
briochh, line 1450: yeeehaa!

briochh (approved):
Some magic to get xy for unstruct... onus on user to provide this definition. Is there a format that we want to support?

jtwhite79:
Def on the user for unstructured xy info - that's why I punted to just using a dictionary of node number xy pairs. That gsf class that @rosskush provided reads this info from a semi standard file that is used with mfusg.   I'm not sure about mf6 unstructured - @mnfienen has an unstructured mf6 freyberg model we could experiment with...maybe flopy has this info?

langevin-usgs:
The modelgrid object in flopy, which is used for mf5, usg, and mf6 models has subclasses for regular, disv, and disu grids.  If there is any way you can build off that, I think it would be helpful, even if it meant those classes needed tweaks.  We now have generalized plotting for unstructured grids and mfusg, including cross section plotting, etc.

jtwhite79:
Thanks @langevin-usgs !  Im not expert witih unstructured grids so any help is appreciated.  We use x-y info about the nodes for pilot points and for geostat simulation if using grid (well, node) scale pars.  Reading thru the mf6 docs, it looks like unstructured in mf6 dispenses with the layer concept (compared to usg) - is this right?  If so, then I think that means we need 3D geostats to support either pilot points or node-scale geostat simulation.  So we can support unstructured mf6 as it is now except for those two cases.  There are plans underway to develop a c++ geostats engine that will support n-d estimation and simulation, so Im hoping we can get this use case covered with that (would be sweet to do 3-D geostat simulation on an unstructured mf6 model!)

langevin-usgs:
So there are a few different flavors of unstructured grids that are relevant to this discussion.  The way we store cell spatial information is through "verts" and "iverts".  verts is just a list of x, y points and iverts is a list that contains a list of vertices for each cell.  flopy relies on this for all its spatial operations.

With this in mind, here are a few points:

1.  The easiest is case is when model uses the same grid for each layer.  In flopy, this is the case for a StructuredGrid, a VertexGrid, and possibly an UnstructuredGrid.  

2.  We use the variable NCPL to denote number of cells per layer.  NCPL is typically a list of length nlay.  For an unstructured grid, if len(iverts) == ncpl[0], then we know that each layer of the unstructured grid is the same, and we don't have to store vertices for all layers.  mfusg allows this type of grid in its disu package.  The disu package in mf6 knows nothing about layers, so the flopy modelgrid must be set up correctly for plotting.

3.  A next type of unstructured grid is a "layered" unstructured grid.  In this case, the model consists of layers, but each layer may have a different grid.  This can only be represented with an UnstructuredGrid in flopy and with the disu package in mf6.  But this is a nice type of grid in that pilot points could still be handled in 2d.

4.  The most complicated grid a fully unstructured grid in which there really is no layering that makes sense.  We can sort of handle this in flopy with the UnstructuredGrid by storing unique vertices and vertex lists for each cell.  In this case, you will need more complicated 3d geostats for interpolation.  

Hope that helps!

jtwhite79:
That does help - thanks bru!

Issue #238, opened 2021-05-05
Observation setup can be slow


briochh:
I have noticed that setting up observations can be (VERY) slow when the model output file contains many columns (ie, the number of characters in the line heads towards 50000). 

I have an example, from a typical MODFLOW6 output (head in every cell -- so heaps of columns) that is taking almost 1 hour to process. 

Most of this seems to be taken up in the inschek routine in  `pst_utils.try_process_output_file()` . 

Much of the slow down relates to repeated slicing of BIG strings (lines of output file), each slice call (e.g. `line[cursor_pos:]`) copies the string and therefore is relatively inefficient. 

Just putting together a PR to try and speed this up but thought I would drop and issue here in case anyone else hits this.

PR #239, opened 2021-05-05
Speed up processing of output files with lots of columns

briochh:
Following on from #238. Trying to speed up processing output files (adding observations with PstFrom()).

Among other attempts for efficiency this PR aims to reduce and speed up the calls to slice long output files lines (e.g. `line[cursor_pos:]`). Using `line.find(str, cursor_pos)` where appropriate and a `maxsearch` length to reduce the length of copied string when slicing. 

Added to pst_from_test to test a big file. Would be nice to add a timeout to this somehow -- but I am not sure how to do that.

### DIFF
--- a/autotest/pst_from_tests.py
+++ b/autotest/pst_from_tests.py
@@ -3018,7 +3018,6 @@ def usg_freyberg_test():
 
 def mf6_add_various_obs_test():
     import flopy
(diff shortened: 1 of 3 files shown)


jtwhite79 (approved):
Looks good B. Nice one.

PR #249, opened 2021-06-24
weightless metrics for goodness of fit

mnfienen:
Added a metrics package and tests for goodness of fit metrics without weights. options to group by obs groups and to include/exclude 0 weighted obs. Also supports RES files and ensembles (ensembles require a PST file as well). Can add functions through a dictionary called `ALLMETRICS` in `pyemu/utils/metrics.py`.


wkitlasten:
Cool!

PR #252, opened 2021-07-06
ineq support in phi calcs, cleanup for upstream pandas and numpy changes

jtwhite79:


pyemu/utils/geostats.py
```diff
@@ -1000,8 +1001,10 @@ def calc_factors_grid(
 
                 dfs.append(df)
                 if var_filename is not None:
+                    #a = np.array([float(str(i)) for i in df.err_var],dtype=np.float).reshape(x.shape)
+                    #a = df.err_var.values.reshape(x.shape)
                     a = df.err_var.values.reshape(x.shape)
-                    na_idx = np.isfinite(a)
+                    na_idx = ~np.isnan(a)
```
briochh, line 1007: are you happy that `a` with never contain `inf` or `-inf`?

jtwhite79: Ah yeah - should go back to infinite.  That was a leftover from the dig

briochh:
bit of No.8 wire on this one @jtwhite79. Nice.

PR #258, opened 2021-07-09
removing negative Kriging weights in factors

mnfienen:
Following this Deutsch (1996) paper: https://doi.org/10.1016/0098-3004(96)00005-2 removing negative Kriging weights which can give nonphysical results if disproportionately large parameter values occur at pilot point locations corresponding with negative weights. 

Made this the default for `calc_factors` the the `OrdinaryKrige` object of `geostats.py`. 

@jtwhite79 - I have a couple questions about the pathing in the autotests before making a good test. I will not, though, that the Kriging factors for Freyberg include some negative values, so we can test with and without the correction.

The correction is made just before the dataframe of factors etc. get assigned to `self.interp_data` in the `OrdinaryKrige` object.


mnfienen:
incorporated a test into `ok_test`

PR #275, opened 2021-08-25
few more additionals to rename functions

jtwhite79:


briochh:
should be covered in #276. Closing for now.

PR #276, opened 2021-09-03
Feat tiedwarnings

jtwhite79:
added more traps for busted tied parameter linking (always hard!).  swapped `PyemuWarning` to inherit from `Warning`, made some really bad issues (like dup par/obs names) errors now when trying to write a `Pst` instance.


briochh:
This one incorporates #275 right? good to go?

jtwhite79:
Yes - sorry forgot that other pr was languishing...

PR #285, opened 2021-09-08
Feat pf operator

jtwhite79:
added an optional `operator` arg to `PstFrom.add_parameters()` to allow non-direct parameters to have an additive action instead of a multiplicative action.  This also required allowing an `initial_value` to `PstFrom.add_parameters()` so that for additive pars, the default initial value is 0 instead of 1.  Added some tests to cover this but its woven in pretty deep.  Also I realize that `par_style="multiplier` is confounded by `operator="+"`.  Not sure if its better to go with `operator` or add a new `par_style`...maybe a new `par_style` to help with the better par naming (#280)?  @briochh thoughts?


briochh (approved):
Seems logical to me. Will be nice to weave this into the parnmes eventually, nice and tidy like.

PR #288, opened 2021-09-22
trying to use numpy over pandas in factor calcs

briochh:
Hoping for some speed up. Just bypassing some expensive pandas lookups (low hanging fruit).

Some minor gains (~20%) in test but most of the work is being done in the interp calc and LinAlg steps (especially for the larger solves).


jtwhite79 (approved):
nice B

Issue #278, opened 2021-09-05
par_type="zone" looking for i,j?


wkitlasten:
I am adding zone multipliers with the following command. The file is an array. It issues a warning that "position of i and j in index_cols not specified." Could be this warning is just out of place... but I could also be missing something?

```
for k in range(nlay):
    pf.add_parameters(filenames=k_files[k],par_type="zone",zone_array=gclass[k],
                      par_name_base=pargp,pargp=pargp,
                      upper_bound=100.,lower_bound=0.01,
                      ult_ubound=1000,ult_lbound=1.0e-6)
```

select output:
```2021-09-05 14:44:10.432157 starting: adding zone type multiplier style parameters for file(s) ['wairau_240_15.k_layer9.dat']
2021-09-05 14:44:10.432157 starting: loading array c:\wairau\wairau_240_15_temp\wairau_240_15.k_layer9.dat
2021-09-05 14:44:10.616339 finished: loading array c:\wairau\wairau_240_15_temp\wairau_240_15.k_layer9.dat took: 0:00:00.184182
2021-09-05 14:44:10.616339 loaded array '..\zmodels\20210904_simulation\wairau_240_15_temp_pst_from\wairau_240_15.k_layer9.dat' of shape (415, 498)
2021-09-05 14:44:10.886829 starting: writing array-based template file 'c:\wairau\wairau_240_15_temp\k_layer9_inst0_zone.csv.tpl'
2021-09-05 14:44:10.886829 starting: writing template file c:\wairau\wairau_240_15_temp\k_layer9_inst0_zone.csv.tpl for ['k_layer9_inst:0']
**2021-09-05 14:44:10.898776 WARNING: get_xy() warning: position of i and j in index_cols not specified, assume (i,j) are final two entries in index_cols.**
2021-09-05 14:44:11.478749 finished: writing template file c:\wairau\wairau_240_15_temp\k_layer9_inst0_zone.csv.tpl for ['k_layer9_inst:0'] took: 0:00:00.591920
2021-09-05 14:44:11.479746 finished: writing array-based template file 'c:\wairau\wairau_240_15_temp\k_layer9_inst0_zone.csv.tpl' took: 0:00:00.592917
2021-09-05 14:44:11.479746 saving zone array c:\wairau\wairau_240_15_temp\k_layer9_inst0_zone.csv.zone for tpl file c:\wairau\wairau_240_15_temp\k_layer9_inst0_zone.csv.tpl
[...]

Issue #277, opened 2021-09-05
Interpolation for the whole gird, or just active domain?


wkitlasten:
I am generating pilotpoint parameters for a fairly big model (400x500) with 15 layers. The active domain gradually reduces to about 500 cells by the last couple layers, so I end up with fewer pilot points in each layer (totally cool and expected). However, the "interp point loop" takes a while and appears to be interpolating for the entire grid (200,000 cells) for each layer. Is that expected? Would it be more efficient to interpolate only for the active domain... or just add unnecessary complexity?

My command:
```
for k in range(nlays):
    pf.add_parameters(filenames=k_file[k],par_type="pilotpoints",
                      use_pp_zones=True, zone_array=idom[k],
                      par_name_base=pargp, pargp=pargp,
                      upper_bound=10., lower_bound=0.1,
                      pp_space=4, geostruct=grid_gs)
```
Selected block of output:
```
2021-09-05 15:37:37.726603 finished: calculating factors for pargp=k_layer8_inst:1 took: 0:03:58.925442
2021-09-05 15:37:37.726603 starting: writing array-based template file 'c:\wairau\wairau_240_15_temp\k_layer8_inst1pp.dat.tpl'
2021-09-05 15:37:37.742236 saving zone array c:\wairau\wairau_240_15_temp\k_layer8_inst1pp.dat.zone for tpl file c:\wairau\wairau_240_15_temp\k_layer8_inst1pp.dat.tpl
2021-09-05 15:37:37.820810 finished: adding pilotpoints type multiplier style parameters for file(s) ['wairau_240_15.k_layer8.dat'] took: 0:04:01.335291
2021-09-05 15:37:37.852030 starting: adding pilotpoints type multiplier style parameters for file(s) ['wairau_240_15.k_layer9.dat']
2021-09-05 15:37:37.852030 starting: loading array c:\wairau\wairau_240_15_temp\wairau_240_15.k_layer9.dat
2021-09-05 15:37:38.005505 finished: loading array c:\wairau\wairau_240_15_temp\wairau_240_15.k_layer9.dat took: 0:00:00.153475
2021-09-05 15:37:38.005505 loaded array '..\zmodels\20210904_simulation\wairau_240_15_temp_pst_from\wairau_240_15.k_layer9.dat' of shape (415, 498)
2021-09-05 15:37:38.180174 starting: writing array-based template file 'c:\wairau\wairau_240_15_temp\k_layer9_inst1_pilotpoints.csv.tpl'
[...]

PR #299, opened 2021-10-15
Refactoring factor calcs

briochh:
Refactoring calc factors in pilot point construction to better handle zones, especially the case there `zone_array` passed to `PstFrom().add_parameters()` defines active domain. Now out of zone model cells are excluded from the interpolation loop. This should help with #277. 
Also propagating correct behaviour for when `use_ibound_zones=False`/`use_pp_zones=False`.
And cleaning up some misleading warnings as per #278


briochh:
I guess it really depends how much of the grid is active. @wkitlasten had calcs speed up from ~270 s to ~90 s with just over 1/3 cells active (although seems to be os dependent). We could pos look at a work around for those mp.Manager lists as I think this is where most of the bottleneck is. Should be poss to spilt up chunks before spawning each worker.

PR #318, opened 2022-01-20
Refactoring shortnames option

briochh:
Proposing to "simplify" setting up parameters and observations with `pst_from` and `longnames=False`. Parameter and observations names with will default to `p{x}` and `o{x}`, respectively, where `x` is effectively a counter. Two additional files will be written to `map` these names to their `longname` counterparts allowing the parameter and observation metadata to still be stored and used in processing (while maintaining better support for classic PEST and PEST UTILS). 

Using the `pyemu.Pst()` function to load in the pest control file will attempt to populate the metadata columns of `parameter_data` and `observation_data` with the metadata stored in these helper files.

These changes do actually simplify the general par and obs naming methods, however, obviously they may break peoples current `longnames=False` setups. Happy to leave on a branch for now if folks want to test more thoroughly.


briochh:
somewhat related to #280

PR #321, opened 2022-02-07
Feat rename speedup

briochh:


jtwhite79 (approved):
Looks good B!  I like that the ~mp.Pool()` pattern b/c it passes exceptions from the child threads back the parent.  Thanks for covering the `append()` stuff too.

PR #325, opened 2022-02-16
Hotfix ineq reweight

briochh:
Noticed that the weight adjustment methods were not quite working where inequality obs were specified.

Obs that won't contribute to the objective function in PESTPP are now filtered out in the weight adjustment.

Also made change in adjust_weights_discrep method but this could stop the resultant phi == number of non zero obs. @jtwhite79 thoughts on this?


jtwhite79:
I think you have it right @briochh - for adjust discrep, the phi should equal nnz obs minus any satisfied ineq obs.

PR #341, opened 2022-04-23
Trying to speed up draws 

briochh:
Rather than attempt to draw a full drag cov after drawing specsim pars (which can had significant overhead where npar its huge), now only drawing pars remaining after specsim draws.

Also chopped out some apply functions that we can get away without -- also can give a solid speedup when npar is sill big.


jtwhite79 (approved):

jtwhite79:
@briochh is this settled and ready to accept?

Issue #352, opened 2022-06-24
Accesing phi and res attributes of PEST control file with cov mat regularisation


rhugman:
Hi,

When reading a PEST(_HP) control file that has covariance matrix file assigned for regularisation of a "regul..." observation group, and then accessing the Pst.phi or Pst.res attributes returns the error: 
`ParserError: Error tokenizing data. C error: Expected 11 fields in line 11675, saw 12`

I assume this is because PEST_HP writes `Cov. Mat.` to the *.res file , which then gets parsed as 2 columns?:

```
'Name                 Group          Measured         Modelled         Residual         Weight       Weight*Measured    Weight*Modelled  Weight*Residual  Measurement_sd   Natural_weight
kx1:1                regulkx1      0.3010300        0.3010300        4.3360196E-09     Cov. Mat.           na               na               na               na               na
```

jtwhite79:
That looks to be the case.  Its hard to parse whitespace delim files when where are extra whitespaces unless they are quoted...not sure what we can do in this case...maybe a fnx to clean up a res file?  Maybe pandas has some ways to help with this?

rhugman:
Specifying `usecols=header` to  line 255 in `pst_utils.py` gets around it. 

```
    res_df = pd.read_csv(
        f, header=None, names=header, sep=r"\s+", converters=converters, usecols=header )

```

rhugman:
Should be OK; I think it will only lose the “extra” column caused by the space in Cov. Mat.. That column is full of NaNs anyway…I’ll test and submit a Pr tmrw.

PR #353, opened 2022-06-26
PR: read res with CovMat regul obs groups

rhugman:
Load Pst with a *.res file that contains rows for CovMat regularization observation groups.
- loads res file, only keeping columns that match headers
- strips "Cov.", "Mat." and "na" strings
- sets cols to float

### DIFF
--- a/pyemu/pst/pst_utils.py
+++ b/pyemu/pst/pst_utils.py
@@ -253,8 +253,12 @@ def read_resfile(resfile):
             header = line.lower().strip().split()
             break
     res_df = pd.read_csv(
-        f, header=None, names=header, sep=r"\s+", converters=converters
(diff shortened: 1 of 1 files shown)


elwan3:
@rhugman 
Brilliant! I had the same issue and thanks to you, it is fixed now.

briochh:
Thanks for this @rhugman, should help address your issue #352. 

Do we have any tests for resfiles with this formatting? Might be worth adding one to help catch this case in the future. Perhaps in `autotest\pst_test.py` or `pst_test_2.py`.

rhugman:
Thanks @briochh. I've added a test based on the henry problem. Let me know if this is OK.

briochh:
Nice one @rhugman, thanks for that. Looking forward to the next PR!

PR #359, opened 2022-08-14
Clean-up files

mwtoews:
Most of these are files that should normally be ignored with `.gitignore`, but seem to have slipped in at some point. A bulk of these files are from an old build of documentation, which is currently done with RTD elsewhere. Other are flopy check files.

Is versioneer actually used in this project? It is partially auto-documented [here](https://pyemu.readthedocs.io/en/latest/autoapi/pyemu/_version/index.html). If it isn't, I have some follow-ups which could clean it up.


jtwhite79:
@mwtoews Im not sure the status of that versioneer stuff.  Feel free to improve/remove....

PR #363, opened 2022-08-25
Remove versioneer, use pyemu/__init__.py for version

mwtoews:
This is a follow-up on #359 to remove versioneer that was introduced with 83b627682c705348607b67cd815e3d1e09f0b7b0

While versioneer is fine, it only works on projects that have a single development branch (usually master or main) with tags. Since this repo has develop with one tag, thus is does not work, e.g. with the develop branch, these versions are different:
```bash
$ python -c "from pyemu import __version__; print(__version__)"
1.0.0+602.g99ad2d3c
$ python setup.py --version
1.2.0
```
With this PR, it is expected to maintain the version in one place: `pyemu/__init__.py`. The develop branch should have something like:
```
__version__ = "1.3.dev0"
```
and when merged to master for the next release, it should be modified to:
```
__version__ = "1.3"
```
----
I have another follow-up PR nearly ready that improves the project metadata and other version handling (which is why I didn't modify setup.py)


mwtoews:
Hmm, perhaps I've bungled up my local git repo. It seems that the tags and versioneer are actually in good condition. I'll need to rethink this one.

PR #370, opened 2022-09-24
GH actions fix

jtwhite79:
not sure what changed with GH actions but heaps of tests are failing b/c of file permission issues on windows...


briochh:
Urg. I have occasionally had to call rmtree in a try and try again call on windows.
try:
    rmtree()
except PermissionBlah:
    Sleep(1)
    rmtree()

Think I have also done this before for a mkdir after a remove. Windows seems to hold on to those locks even after the python function return.

jtwhite79:
It was esp happening with .hds files so I dont know if flopy was hanging on to those file extensions...but it just started happening and I cant repro it locally so Im guessing something on GH actions changed...

PR #387, opened 2022-12-30
feat_disv_spatialreference

rhugman-intera:
Not the most elegant, but functional, fixes to enable spatial reference using MF6 DISV grids. As far as my local tests show, I don't seem to have broken anything else /fingers crossed. 

Added a vertex version of freyberg for testing. Adapted from [here](https://github.com/DOI-USGS/python-for-hydrology/blob/main/notebooks/part1_flopy/solutions/06-Project-quadtree.ipynb)

Added a pst_from test with vertex model.

(BTW, looks like it wouldn't be too complicated to do similar for DISU.)


briochh:
Sorry @rhugman-intera, I missed this. Thanks for your effort on this.

It looks like the test is failing. Possibly the par change checks are causing the forward model to break? If you just want to check that the par changes are having an effect (i.e. you don't actually need to run the model) you could use `pyemu.helpers.apply_list_and_array_pars()` (run from inside the model directory so need to cd first).

As a side note, I have been refactoring the CI to use pytest, this means that the style of some of the test have changed (still, probably not as efficient as they could be).  That is ready to merge now. You'll need to resync your branch after. Then see if you can modify your test to get it running in that new CI. I can help with this, if needed.

rhugman-intera:
Thanks @briochh. Merged and changed to `apply_list_and_array_pars()` as suggested.

autotest/pst_from_tests.py
```diff
@@ -4576,6 +4577,184 @@ def shortname_conversion_test(tmp_path):
...
+    par_org = par.copy()
+    #check gr pars
+    for zone in par.zone.unique():
+        par.loc[(par.zone==zone) & (par.ptype=='gr'), 'parval1'] = float(zone)
+        par.loc[(par.zone==zone) & (par.ptype=='gr'), 'parlbnd'] = float(zone)-0.1
+        par.loc[(par.zone==zone) & (par.ptype=='gr'), 'parubnd'] = float(zone)+0.1
+
+    # write model input files
+    pst.write_input_files(template_ws)
+    pyemu.os_utils.run(r'python forward_run.py', cwd=template_ws)
```
briochh, line 4718: It looks like the mf6 run within forward_run.py is failing. Possibly the changes to parval1, above, move values beyond stable values for the forward run? Does it run locally with these values?

rhugman-intera: Hey @briochh, sorry about the hassle...the refactored `pst_from_tests.py` pytests run locally...I've added a few fixes to some other issues that were failing locally...but none of them seem to be related to the failing autotest tests.

briochh: Looks like you got it @rhugman-intera. Good job dealing with the moving goal posts.

rhugman-intera:
Thanks @briochh. I've added some functionality to place pilot points for a vertex grid on my fork. It relies on `shapely`. Is this a problem? (I assume you are trying to avoid relying on other dependencies).

PR #393, opened 2023-01-17
Feat vertex

rhugman-intera:
+ pilot points for vertex grids


mwtoews:
also add "shapely" to: https://github.com/pypest/pyemu/blob/bd182426727a8a211a3894c28a485c552b19164b/pyproject.toml#L30-L35

There are some potential shapely 2.0 performance enhancements using numpy arrays of geometries, but these can happen later.

briochh:
thanks @rhugman-intera

PR #394, opened 2023-01-17
Upgrade versioneer, clean-up setup.cfg and MANIFEST.in

mwtoews:
This PR does the following:

- Upgrade the vendored versioneer, which uses the configuration from pyproject.toml, therefore clean up setup.cfg.
- Also MANIFEST.in is not necessary, as these files are included with the sdist.
- Repeat "name" in setup.py to [avoid a GitHub bug](https://github.com/orgs/community/discussions/6456), which could happen when `master` is synced to `develop`


briochh:
thanks @mwtoews

PR #415, opened 2023-04-04
fix(Pst.adjust_weights): RuntimeError: dictionary changed size during…

### LINKED ISSUE #414: RuntimeError when adjusting weights with some groups zero-weighted
This error can be reproduced with the following code (from the `pst_tests::reweight_test` function):

```python
import os
import numpy as np
from pyemu import Pst
pst_dir = os.path.join("pst")
p = Pst(os.path.join(pst_dir, "pest.pst"))
obsgrp_dict = {"pred": 0., "head": 1.0, "conc": 1.0}
p.adjust_weights(obsgrp_dict=obsgrp_dict)
```
which results in:
```
Traceback (most recent call last):
  File "<string>", line 2, in <module>
  File "/Users/aleaf/Documents/GitHub/pyemu/pyemu/pst/pst_handler.py", line 2588, in adjust_weights
    for grp, contrib in obsgrp_dict.items():
RuntimeError: dictionary changed size during iteration
```

aleaf:
… iteration when some groups in obsgrp_dict are set to zero weight. Fixes #414

### DIFF
--- a/pyemu/pst/pst_handler.py
+++ b/pyemu/pst/pst_handler.py
@@ -2587,11 +2587,14 @@ class Pst(object):
         if obsgrp_dict is not None:
             obs = self.observation_data
             # first zero-weight all obs in groups specified to have 0 contrib to phi
+            zero_weight_groups = []
             for grp, contrib in obsgrp_dict.items():
                 if contrib==0:
                     obs.loc[obs.obgnme == grp, "weight"] = 0.0
                     # drop zero- contribution groups
-                    del obsgrp_dict[grp]
+                    zero_weight_groups.append(grp)
+            obsgrp_dict = {k: v for k, v in obsgrp_dict.items() 
+                           if k not in zero_weight_groups}
             # reset groups with all zero weights
             for grp in obsgrp_dict.keys():
                 if obs.loc[obs.obgnme == grp, "weight"].sum() == 0.0:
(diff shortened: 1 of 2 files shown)


pyemu/pst/pst_handler.py
```diff
@@ -2585,11 +2585,14 @@ def adjust_weights(self, obs_dict=None, obsgrp_dict=None):
...
                 if contrib==0:
                     obs.loc[obs.obgnme == grp, "weight"] = 0.0
                     # drop zero- contribution groups
-                    del obsgrp_dict[grp]
+                    zero_weight_groups.append(grp)
+            obsgrp_dict = {k: v for k, v in obsgrp_dict.items() 
+                           if k not in zero_weight_groups}
             # reset groups with all zero weights
             for grp in obsgrp_dict.keys():
                 if obs.loc[obs.obgnme == grp, "weight"].sum() == 0.0:
```
briochh, line 2598: I wonder if we can combine these loops in to one (and avoid that runtime dict size change error) by looping over `list(obsgrp_dict.keys())`.
So:
```
for grp in list(obsgrp_dict.keys()):
    if obsgrp_dict[grp]==0:
        obs.loc[obs.obgnme==grp,"weight"]=0
        del obsgrp_dict[grp]
    elif obs.loc[obs.obgnme==grp,"weight"].sum()==0:
        obs.loc[obs.obgnme==grp,"weight"]=1
```

briochh (commented):
Thanks for this @aleaf, and thanks for dropping in some more testing lines. 

I think it looks like it should work but have put in a comment about a possible mod (more to that bit of the code in generally than to your fix).

Might be worth merging develop back into your branch, that should chase out those Pandas 2.0 related fails.

briochh:
Thanks @aleaf

PR #416, opened 2023-04-09
fixes for pandas 2

jtwhite79:
ran out of steam on one for the res 1-to-1 plot tests, otherwise, things seem to be at least passing tests now...


pyemu/pst/pst_handler.py
briochh, line 1: Looks like a lot of the Pandas2 issues can be traced back to this 
`df.loc[:, col] = df.loc[:, col].apply(str).apply(converters[col])` assignment. I have run a few tests and it looks like RHS values are now cast to LHS dtype when using this type of assignment. Not sure if this is a bug in Pandas or not.
Even `df.loc[:, col] = df.loc[:, col].astype(int)` will not modify the dtype of `df.loc[:, col]`. As a result this conversion line was not doing anything and our dtypes were remaining as object type causing downstream np.ufuncs to fail.

briochh (commented):
@jtwhite79, thanks for chasing this through. I have done some more hunting and it looks like the issues can be traced back to a change in how `df.loc[:, col]=RHS` assignments are working in Pandas 2.0 (dtype changes are not happening as we might think/need).

I have made some changes locally that hopefully mean that we can strip back some of these mods. I'll push to this commit.

briochh:
https://github.com/pandas-dev/pandas/issues/52612#issue-1663387395

Issue #418, opened 2023-04-11
error from pyemu.en.ParameterEnsemble.from_mixed_draws


RyanConway91:
I am getting 

```
Cell In[3], line 1
    dists = pyemu.en.ParameterEnsemble.from_mixed_draws(pst, how_dict, default='gaussian',

  File ~\AppData\Local\miniforge3\envs\gwmodel_2023\Lib\site-packages\pyemu\en.py:1229 in from_mixed_draws
    df.loc[pe.index, pe.columns] = pe

  File ~\AppData\Local\miniforge3\envs\gwmodel_2023\Lib\site-packages\pandas\core\indexing.py:818 in __setitem__
    iloc._setitem_with_indexer(indexer, value, self.name)

  File ~\AppData\Local\miniforge3\envs\gwmodel_2023\Lib\site-packages\pandas\core\indexing.py:1795 in _setitem_with_indexer
    self._setitem_with_indexer_split_path(indexer, value, name)

  File ~\AppData\Local\miniforge3\envs\gwmodel_2023\Lib\site-packages\pandas\core\indexing.py:1831 in _setitem_with_indexer_split_path
    self._setitem_with_indexer_frame_value(indexer, value, name)

  File ~\AppData\Local\miniforge3\envs\gwmodel_2023\Lib\site-packages\pandas\core\indexing.py:1947 in _setitem_with_indexer_frame_value
    if item in value:

TypeError: argument of type 'ParameterEnsemble' is not iterable
```
with:

```
dists = pyemu.en.ParameterEnsemble.from_mixed_draws(pst, how_dict, default='gaussian', 
   num_reals=1000, cov=None, sigma_range=4, enforce_bounds=True, partial=False, fill=True)
```

how_dict does not include all the parameters (there are fixes and ties I want to enforce), but I don't think thats the issues

jtwhite79:
Hey Ryan - I can't reproduce this with the testing files that we use.  Can you post the .pst file?

RyanConway91:
pst file below. Thanks!
pcf
* control data
 restart estimation
#NPAR    NOBS      NPARGP NPRIOR NOBSGP [MAXCOMPDIM][DERZEROLIM]
    72 13    16           0      7
#NTPLFLE NINSFLE PRECIS  DPOINT [NUMCOM][JACFILE] [MESSFILE] [OBSREREF]
  14     5       single  point  1        0         0
[...]

PR #421, opened 2023-04-18
Hotfix mixeddraws

jtwhite79:
fix for #418 and also fix for using floats as index columns in list-style pars in`PstFrom`


pyemu/utils/helpers.py
```diff
@@ -2154,6 +2169,8 @@ def _process_list_file(model_file, df):
             common_idx = (
                 new_df.index.intersection(mlts.index).sort_values().drop_duplicates()
             )
+            if common_idx.shape[0] == 0:
```
briochh, line 2172: Ee might be a able to do better here. Perhaps we can pass the number of expected alignments to that tracker file and then check against that here? Testing something at my end now, but lets go with this for them being.

briochh (approved):
Thanks for wading through this, at least the mixed draws was as easy fix!

pyemu/utils/helpers.py
jtwhite79: agreed - explicitly tracking the original alignment would be much safer...

PR #422, opened 2023-04-19
Adding check process to alignment at apply pars time

briochh:
Adding additional data store during add pars and checking against as apply pars time. 

Hopefully should catch misalignment of`index_cols` when applying pars.

Caveat: might cause unexpected failures... may need refining.


pyemu/utils/pst_from.py
```diff
@@ -968,7 +968,7 @@ def _par_prep(
                         sep = " "
                         if rel_filepath.suffix.lower() == ".csv":
                             sep = ","
-                if df.columns.is_integer():
+                if pd.api.types.is_integer_dtype(df.columns):  # df.columns.is_integer(): # really!???
```
jtwhite79, line 971: yikes!

jtwhite79 (approved):
Looks good to me.  I think having this check will help identify bugs sooner...

pyemu/utils/pst_from.py
briochh: vile ey?

briochh: https://github.com/pandas-dev/pandas/issues/50042

Issue #433, opened 2023-05-31
Restarting pestpp-glm from pyemu


Paszka1:
Is it possible to restart an accidentally interupted parallel pestpp-glm inversion from pyemu, or not? I.e., I would like to avoid restarting from the command line as I'm not sure how to reuse the existing workers.

mnfienen:
Hey @Paszka1 - I'm afraid if you have to restart an interrupted run, you have to restart the workers from scratch.

I will experiment a bit with adding the option to include the restart `/r` string to the command call in `start_workers()` though

mnfienen:
ok @Paszka1  - there's a pull request in on the develop branch that should allow restarting using `start_workers`. It will require specifying `restart=True` and `reuse_master=True`. You can see the status of the pull request on the link above and give it a try.

PR #434, opened 2023-05-31
(feat) adding restart option to os_utils.start_workers()

mnfienen:
Simple `restart` flag added to `os_utils.start_workers()` that adds `/r` to the call string only for the master if `restart==True`. Else no difference in behavior (default `restart=False`.

I couldn't find a good test to include this in within the autotest suite, but I did confirm it works on a simple problem and restarts properly.

Addressing #433


briochh:
Sorry @mnfienen, backlogged. I don't see any reason why this wouldn't work, thoughts @jtwhite79?

mnfienen:
I went ahead and accepted this one....

PR #446, opened 2023-07-19
(feat) added get_phi_vector_noise_obs() method 

mnfienen:
for ObservationEnsemble and a test. There is a test in there as well. Have a look.

### DIFF
--- a/pyemu/en.py
+++ b/pyemu/en.py
@@ -811,8 +811,8 @@ class ObservationEnsemble(Ensemble):
         """
         cols = self._df.columns
         pst = self.pst
-        weights = self.pst.observation_data.loc[cols, "weight"]
-        obsval = self.pst.observation_data.loc[cols, "obsval"]
+        weights = pst.observation_data.loc[cols, "weight"]
+        obsval = pst.observation_data.loc[cols, "obsval"]
         phi_vec = []
         for idx in self._df.index.values:
             simval = self._df.loc[idx, cols]
--- a/pyemu/en.py
+++ b/pyemu/en.py
@@ -820,6 +820,63 @@ class ObservationEnsemble(Ensemble):
             phi_vec.append(phi)
         return pd.Series(data=phi_vec, index=self.index)
 
+    def get_phi_vector_noise_obs(self, noise_obs_filename=None):
+        """vector of L2 norm (phi) for the realizations (rows) of `Ensemble`. This
+            differs from `phi_vector` in that phi is calculated for each ensemble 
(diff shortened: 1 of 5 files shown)


mnfienen:
Please don't merge yet - I'm working on some refactoring

pyemu/en.py
```diff
@@ -811,15 +811,72 @@ def phi_vector(self):
...
+                    f"Could not open {noise_obs_filename}"
+                )
+        # make sure the index is all string (since the real)
+        noise_obs.index = [str(i) for i in noise_obs.index]
+        
+        # now make the calculations
+        cols = self._df.columns
+        weights = self.pst.observation_data.loc[cols, "weight"]
+        phi_vec = []
+        for idx in self._df.index.values:
```
briochh, line 872: wonder if you could get away with:
`phi_vec = (((self._df-noise_obs) * weights)**2).dropna().sum(axis=1)`
pandas should align nicely on index and columns right?

mnfienen: maybe, but @jtwhite79  pointed out that we need to handle inequality constraints, so gotta sum by groups and handle conditions about violation of inequality constraints. This led to some refactoring which may be close to done. If we get it working, I still like the idea of de-looping but not sure we can

briochh: ah yea, good call. I think should be poss to work with a residual ensemble `res = (self._df-noise_obs)` and then zero out the residual that are the wrong sign for obs that are in inequality groups. `res.loc[:, ltobs] = res.loc[:, ltobs].clip(0)`

mnfienen: that sounds like a good idea. I might try and look tomorrow. If you get a chance, maybe check out the latest commit I pushed. I think it's working, but too slow to be practical. It worked for a simple case without noise, but the noise obs example is a bigger file and def too slow. Late here but I can look tomorrow or Friday again

mnfienen: All the tests are passing now so I think we are good, but the ensemble version is pretty slow. I like your idea @briochh so might try and implement that way

briochh:
did you want to add more to this @mnfienen?

mnfienen:
hey @briochh - I've been away for the month so missed this. I'm tempted to fold it in and experiment with the other option later unless you have time to test it?

briochh:
Welcome back @mnfienen . Yeah go for the merge. We can look into testing some improves later from dev branch.

mnfienen:
Right on @briochh  - merging

PR #457, opened 2023-08-31
feat(get-pestpp): add get-pestpp command line interface

jdhughes-usgs:


briochh:
@jdhughes-usgs just noting this in here. Looks like the failing tests are related to pandas 2.1 update and pd.read_csv() with delim_whitespace. Looks unrelated to your PR.

jdhughes-usgs:
@briochh There were two columns in one line of one example dataset that was causing the pandas problem. Let me know if this is not a valid fix.

briochh:
@jdhughes-usgs, yeah but I think that the "extra" column was split by a comma. @jtwhite79, might be able to comment on whether the comma is valid in there, think it has always been there. Looks like the `delim_whitespace=True` option now still splits on `,`, which is a bit weird. 

Man, spending a lot of time chasing breaking changes these days!

jdhughes-usgs:
> @jdhughes-usgs, yeah but I think that the "extra" column was split by a comma. @jtwhite79, might be able to comment on whether the comma is valid in there, think it has always been there. Looks like the `delim_whitespace=True` option now still splits on `,`, which is a bit weird.
> 
> Man, spending a lot of time chasing breaking changes these days!

@briochh It was split by a comma. It was the only line with an extra comma. Hopefully @jtwhite79 can let me know what needs to be done.

briochh:
@jdhughes-usgs pin pandas? https://github.com/pandas-dev/pandas/issues/54918

jdhughes-usgs:
@briochh could do that but could be something that bites later on.

PR #459, opened 2023-09-06
ci: update workflow to use get-pestpp cli to get PEST++ executables

jdhughes-usgs:


briochh:
Awesome stuff, thanks @jdhughes-usgs

jdhughes-usgs:
> Awesome stuff, thanks @jdhughes-usgs

No problem. Glad it turned out to be simple.

PR #460, opened 2023-09-06
ci: add GITHUB_TOKEN to bypass rate limit

jdhughes-usgs:


briochh:
Was just looking at this one @jdhughes-usgs. Interesting that the rate limit was only triggered when running the CI on merge. I don't understand the GITHUB_TOKEN usage well enough. Have seen other deployments that use.  
```
with:
    repo-token: ${{ secrets.GITHUB_TOKEN }}
```
assume that it is equiv.

jdhughes-usgs:
I don't know enough about the differences but went with the `env` approach since the GH Actions error was:

```
ValueError: use GITHUB_TOKEN env to bypass rate limit (HTTP Error 403: rate limit exceeded)
```

briochh:
good logic, let's give it a whirl

PR #478, opened 2024-02-18
using nanmax and nanmin on axis limits to deal with missing values in ensembles

wkitlasten:
To allow sending an ensemble with outliers masked by nans without messing up axis limits.


wkitlasten:
faulty

PR #451, opened 2023-08-10
more pandas  fixes

briochh:
Another catch for pandas assignment changes. object types not playing nice with numpy ufuncs.

PR #489, opened 2024-03-20
Unpin Pandas and update CI

### LINKED ISSUE #461: Pandas pinned to < 2.1.0
A few issues with behaviour of pandas 2.1.0 that causes failed tests and a whole string of Future and Deprecation warnings

Need to assess what actually reflects an impending breakage in how we are using pandas, going forwards and what might be temporary pandas "features" as they move towards pyarrow.

briochh:
Pandas was pinned `pandas<2.1.0` to give time to chase some potentially breaking deprecations. This PR removes the pin (tentatively!) after recent PRs attempted to fix some upcoming deprecations  #451 #485 #486 (some more addressed in this PR). Should close #461.

Also updated CI to use latest version for checkout and coveralls actions and added python 3.12 to test suite.


pyproject.toml
```diff
@@ -1,7 +1,7 @@
 [build-system]
 requires = [
     "setuptools >= 61",
-    'tomli; python_version < "3.11"',
+    'tomli; python_version <= "3.12"',
```
mwtoews, line 4: this doesn't need changing, since [tomllib](https://docs.python.org/3/library/tomllib.html) is used in [versioneer.py](https://github.com/pypest/pyemu/blob/d390d5d175ca831abbc6480da8d078107e0f4418/versioneer.py#L37) for py311+

briochh: ah that's right, good catch @mwtoews

Issue #490, opened 2024-03-21
pst_from support for single column parameter files


wkitlasten:
MODFLOW can handle a single column of data with nlay x nrow x ncol entries (e.g., K). pst_from.add_parameters() does not seem to like those files (no index, no header). Doesn't it seem reasonable to support such a file when index_cols=None and use_cols=None?

I will try to address this tomorrow (fair warning!). Anything I might be missing?

briochh:
@wkitlasten this seems like something that should be supported. I would vote for using index_col=None but specifying use_col=0 for the headerless single column list file type. 

The main thing to look out for is that (I think) the presence of a non None index_col is what triggers a lot of assumptions about the input file structure (ie list rather than array style). This could be switched over to use_cols being the trigger and the index that is required for the parameter name being derived from the dataframe index. 

There maybe a bit of chasing it through required. Give a shout if you get stuck.

briochh:
The option here is to support single col inputs as either list-like or array-like.
For **list-like**, we would need to support passing to `add_parameters()` without `index_cols` argument and move the list-like setup trigger to `use_cols`. May just need chasing through methods a bit.
For **array-like**, we would need to [circumvent](https://youtu.be/y2WaVhAF9TE?si=KJzttaH7fPerHRpd)  the automatic `get_xy` trigger.
The array option might actually be easier to implement but for me it feels better to set these up list-like. Obviously supporting spatial definition for inputs like this is tricky and we might need to look at allowing more granular specification of arguments at `add_pars` time to support mixed model input format types (i.e. specification of input specific `get_xy` function/spatial mapping rather than relying on what is on the parent object).

PR #498, opened 2024-04-17
Better preservation of input file precision through add_pars and apply_mults

briochh:
Should result in less precision noise in workflows


briochh:
actually might help with #490

pyemu/utils/helpers.py
```diff
@@ -1689,7 +1689,17 @@ def _process_array_file(model_file, df):
                 lb = float(list(lb_vals.keys())[0])
                 org_arr[org_arr < lb] = lb
 
-    np.savetxt(model_file, np.atleast_2d(org_arr), fmt="%15.6E", delimiter="")
+    try:
+        fmt = df_mf.fmt.iloc[0]
+    except AttributeError:
+        fmt = "%15.6E"
```
jtwhite79, line 1695: I guess we need to keep the 15-6 for old school modflow/mt3d?

briochh: I think with PstFrom it should be ok. The detect 'should' get the right format. We need to is fall-back for the old pstfromflopy method which I don't think stores a 'fmt' (and I didn't fancy opening that box!)

PR #503, opened 2024-05-03
Changes to mat_handler.py

laat0003:
Added a test for diagonal matrix in Matrix.extract that will reset inherited diagonal status to True if the extracted matrix turns out to be diagonal.

Fixed an error with to Cov.to_uncfile where the standard deviations written to the uncfile were not from the diagonal.

Newb at the fork commit pull request stuff so please let me know if I screwed this up :)

### DIFF
--- a/pyemu/mat/mat_handler.py
+++ b/pyemu/mat/mat_handler.py
@@ -1825,6 +1825,10 @@ class Matrix(object):
         if row_names is None and col_names is None:
             raise Exception("Matrix.extract() " + "row_names and col_names both None")
         extract = self.get(row_names, col_names, drop=True)
+        i,j = extract.x.shape
+        if i==j:
+            test = extract.x.reshape(-1)[:-1].reshape(i-1,j+1)
+            extract.isdiagonal = ~np.any(test[:,1:])
         return extract
 
     def get_diagonal_vector(self, col_name="diag"):
@@ -3067,7 +3071,7 @@ class Cov(Matrix):
                 f.write("START STANDARD_DEVIATION\n")
                 for iname, name in enumerate(self.row_names):
                     f.write(
-                        "  {0:20s}  {1:15.6E}\n".format(name, np.sqrt(self.x[iname, 0]))
+                        "  {0:20s}  {1:15.6E}\n".format(name, np.sqrt(self.x[iname, iname]))
                     )
                 f.write("END STANDARD_DEVIATION\n")
                 f.close()


pyemu/mat/mat_handler.py
```diff
@@ -3067,7 +3071,7 @@ def to_uncfile(
                 f.write("START STANDARD_DEVIATION\n")
                 for iname, name in enumerate(self.row_names):
                     f.write(
-                        "  {0:20s}  {1:15.6E}\n".format(name, np.sqrt(self.x[iname, 0]))
+                        "  {0:20s}  {1:15.6E}\n".format(name, np.sqrt(self.x[iname, iname]))
```
jtwhite79, line 3074: I dont think we want to change to expecting two full dimensions here - this is within a conditional that the instance is in fact diagonal and the matrix class stores a diagonal matrix as a vector.

pyemu/mat/mat_handler.py
```diff
@@ -1825,6 +1825,10 @@ def extract(self, row_names=None, col_names=None):
         if row_names is None and col_names is None:
             raise Exception("Matrix.extract() " + "row_names and col_names both None")
         extract = self.get(row_names, col_names, drop=True)
+        i,j = extract.x.shape
+        if i==j:
```
jtwhite79, line 1829: I think the way to test if the extracted matrix is diagonal is that if i == j and np.count_nonzero(self.x) == np.count_nonzero(np.diag(self.x)).  I think that last condition would make sure there are no off diagonals (in the unlikely case where there are zeros on the diagonal)?

pyemu/mat/mat_handler.py
laat0003: Thanks @jtwhite79. These comments helped me clear up where my problem lies. 

I couldn't get the extract method to save a vector. I was extracting from a Cov built with PstFrom.build_prior and then using .to_uncfile on it. The extracted matrix was diagonal, which it had to be because the parameters being extracted were not associated with any geostruct but .to_uncfile couldn't tell that it was diagonal despite writing a diagonal matrix to a file.  My cheat was to force a check for off-diagonal entries on the extracted Cov post-extraction and if none, then reset .isdiagonal property while using the entries along the diagonal of the square matrix for the standard deviations block in the .to_uncfile write. 

At the time I thought that the .isdiagonal method was actually testing if the matrix was diagonal but I only figured out later (after submitting this PR) that it is just accessing the property. 

I think that the reason the extracted Cov wasn't being flagged as diagonal is because of the .get method. When it operates on a Cov, it doesn't actually test for diagonal on the new extracted Cov but just passes on the .isdiagonal status of the original Cov. This forces .to_uncfile to ask for a filename to write the new Cov to because the matrix remains non-diagonal.

I can't use the huge single Cov from PstFrom.build_prior (which kicked off this journey) because PEST_HP complains if a parameter that should be in a standard deviation block is in a Cov. For context, I'm working on a way to streamline going from a pre-existing IES build to an ENSI specfile.

laat0003: I'd like to have another crack at fixing the problem in the .get method - if that is okay with you. If it is better for you if I just report a bug I'll do that instead. Please let me know which you'd prefer. Thanks again for your time.

pyemu/mat/mat_handler.py
laat0003: Please see reply comments on the other change as to why I was doing this. In hindsight it was stoopid to try a band-aid solution that worked for my situation in PR. Thanks for your patience.

pyemu/mat/mat_handler.py
jtwhite79: Please do have another go at it!  I think it might more explicit add a `to_diagonal()` method to the `Cov` class - then you'd just pass the extracted Cov thru it before writing the unc file?  
Also, I think you have to enable the github actions on your pyemu fork so that the tests will run for you (so you can see that your changes are good-to-go as far as the testing goes...and you could even add more tests!)

laat0003: Will do. Thanks for the advice on both Github use and the ideal approach to solving my little problem. I'll see if I can get it sorted today. All the best.

Issue #510, opened 2024-05-17
pe.enforce() does not always update df


nicohiggs:
Ran into a situation where pe.enforce() was not enforcing bounds. After some digging I think it is related to how df.values will sometimes return a copy instead of a view (see MSeifert's answer here https://stackoverflow.com/questions/48201521/will-changes-in-dataframe-values-always-modify-the-values-in-the-data-frame and the issue here https://github.com/pandas-dev/pandas/issues/19219). Digging into _enforce_reset() you can see how if df.values returns a copy then editing the copy will not update the df.

The trouble is that when inspecting the df after calling pf.draw() in my model it appears all my types are the same, so I am not totally sure why in my case _enforce_reset() is making a copy with self._df.values which results in the df not being updated and enforce() not working. I was able to fix this locally by being more explicit in _enforce_reset() and will make a PR.

briochh:
closed by #511 , thanks @nicohiggs

PR #511, opened 2024-05-17
_enforce_reset() to be more explicit

nicohiggs:
This fixes issue #510 

Again I am not 100% why I was getting this issue for one of my models because the dtypes appeared to be fine after drawing ensembles, but it was clear pe.enforce() was not working for me (values outside both the lower and upper bounds not changing) and this fix I made by locally editing _enforce_reset() fixed it. I added a test that fails without this _enforce_reset() update but passes with it.

The vast majority of the time I think the dtypes should all be float64/upcastable to float64 and this will not be an issue. But I think it is a good idea to be more explicit in general anyways as nothing will notify you about this (pandas documentation has a warning now about using values: https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.values.html and in the links in issue #510 they discuss how this in an implementation detail and they might change implementation thus the pandas team aren't changing this behaviour).

### DIFF
--- a/autotest/en_tests.py
+++ b/autotest/en_tests.py
@@ -260,6 +260,7 @@ def enforce_test():
     pe.enforce(how="scale")
 
     pe = pyemu.ParameterEnsemble.from_gaussian_draw(pst, num_reals=num_reals)
+    pe._df["mult1"] = pe._df["mult1"].astype("float")
     pe._df.loc[0,:] += pst.parameter_data.parubnd
     pe.enforce()
     assert (pe._df.loc[0,:] - pst.parameter_data.parubnd).apply(np.abs).sum() == 0.0
@@ -718,7 +719,7 @@ if __name__ == "__main__":
     #deviations_test()
     # as_pyemu_matrix_test()
     # dropna_test()
-    #enforce_test()
+    enforce_test()
     #pnulpar_test()
     # triangular_draw_test()
     # uniform_draw_test()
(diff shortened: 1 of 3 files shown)


briochh:
Hi @nicohiggs, looks like a good catch, we should definitely be passing that np array back to the parent dataframe. At the risk of exposing us more to the vagrancies of pandas changes, I wonder if we could do the truncation/clipping directly on the dataframe from pandas. 
`self._df = self._df.clip(lower=lb, upper=ub, axis=1)` I think that will take the upper and lower bound dictionaries and will avoid the loop and casting to numpy and back again?

Not sure on the failing tests, might need some more digging.

nicohiggs:
I am wondering if you guys use dictionaries as a way to guard against the order of the columns of self._df and self.lbnd and self.ubnd possibly being different? It looks like df.clip accepts a dict but I am doubtful of it using the keys to ensure things line up when out-of-order, I think it likely just extracts the values and uses them like an array. If this is not a concern then I think my latest commit will be fine, if it is a concern though then there is some extra work that could be done to ensure that self.lbnd and self.ubnd have the same order as the columns of self._df. This extra work could be done here, but it potentially should something done beyond/outside of this function.

briochh:
I think @jtwhite79 has a good theory on why these tests are now failing after the change. Hopefully he is baking up a fix. We might ask you to merge that in to this PR once he is done.

jtwhite79:
@nicohiggs I opened a pr on your fork that should fix the tests. I think if accept it, gh should automatically update this pr...

briochh:
Thanks @nicohiggs, good catch 'n' fix!

PR #514, opened 2024-06-07
Extracted most of pst_from.draw() into helpers

wkitlasten:
I was having a hard time drawing ensembles with lots of pars any other way, so I borrowed the bits from pst_from and figured we could make them more accessible via helpers. All tests passed locally, hopefully I didn't break anything!


pyemu/en.py
```diff
@@ -480,7 +480,7 @@ def _gaussian_draw(
                     # cnames = names
                     cnames = []
                     snames = set(names)
-                    for n in cov.names:
+                    for n in cov.col_names:
```
briochh, line 483: I think `.names` should work so long as the `cov` object that is passed to the method is `Cov` type not `Mat`.  Does this not work for you?

pyemu/utils/helpers.py
```diff
@@ -235,6 +235,137 @@ def autocorrelated_draw(pst,struct_dict,time_distance_col="distance",num_reals=1
             full_oe.loc[:, n] = v
     return full_oe
 
+def draw_by_group(pst, num_reals=100, sigma_range=6, use_specsim=False,
+         struct_dict={}, delr=[], delc=[], scale_offset=True, echo=True, logger=False):
+    """Draw a parameter ensemble from the distribution implied by the initial parameter values in the
+    control file and the prior parameter covariance matrix. Previously in pst_from.
```
briochh, line 241: As we don't explicitly pass the prior par cov here, it might be clearer to say "... a prior parameter covariance matrix derived from grouped geostructures" or something like that.

pyemu/utils/helpers.py
```diff
@@ -235,6 +235,137 @@ def autocorrelated_draw(pst,struct_dict,time_distance_col="distance",num_reals=1
...
+
+        If you are using grid-style parameters, please use spectral simulation (`use_specsim=True`)
+
+    """
+    if not logger:
+        logger = pyemu.Logger("draw_by_groups.log", echo=echo)
+    if pst.npar_adj == 0:
+        logger.warn("no adjustable parameters, nothing to draw...")
+        return
+    # precondition {geostruct:{group:df}} dict to {geostruct:[par_dfs]}
```
briochh, line 278: drop comment - this is handled in the `PstFrom._pivot_par_struct_dict()` method (might be possible to abstract this too, to allow more flexibility of the form of that `struct_dict` object).

pyemu/utils/helpers.py
```diff
@@ -235,6 +235,137 @@ def autocorrelated_draw(pst,struct_dict,time_distance_col="distance",num_reals=1
...
+            logger.log(f"Joining specsim and non-specsim pars")
+            exist = gr_par_pe.columns.intersection(pe.columns)
+            pe = pe._df.drop(exist, axis=1)  # specsim par take precedence
+            pe = pd.concat([pe, gr_par_pe], axis=1)
+            pe = pyemu.ParameterEnsemble(pst=pst, df=pe)
+            logger.log(f"Joining specsim and non-specsim pars")
+    else:
+        pe = pyemu.ParameterEnsemble(pst=pst, df=gr_par_pe)
+    logger.log("drawing realizations")
+    return pe.copy()
```
briochh, line 367: note to self: work out why we were returning a copy here (probs a pandas ref vs copy thing).

briochh (changes requested):
Great stuff @wkitlasten. A few minor mods to comments but the code looks like it should work well in the abstracted form...

pyemu/utils/helpers.py
```diff
@@ -235,6 +235,137 @@ def autocorrelated_draw(pst,struct_dict,time_distance_col="distance",num_reals=1
...
+        num_reals (`int`): the number of realizations to draw
+        sigma_range (`int`): number of standard deviations represented by parameter bounds.  Default is 6 (99%
+            confidence).  4 would be approximately 95% confidence bounds
+        use_specsim (`bool`): flag to use spectral simulation for grid-scale pars (highly recommended).
+            Default is False
+        struct_dict (`dict`): struct_dict (`dict`): a dict of GeoStruct (or structure file), and list of
+            pilot point template files pairs. If the values in the dict are
+            `pd.DataFrames`, then they must have an 'x','y', and 'parnme' column.
+            If the filename ends in '.csv', then a pd.DataFrame is loaded,
+            otherwise a pilot points file is loaded.
```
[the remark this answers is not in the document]
wkitlasten: Yes, this is critical. I copied the language straight from "geostatistical_draws" in an effort to be consistent and since that is where struct_dict gets passed. I probably missed something. I agree it could be made more flexible, but I wanted to minimize changes until I was sure the extraction was successful. Also worried about chasing any changes through the rest and breaking something!

pyemu/en.py
wkitlasten: I never would have seen that line if I hadn't gotten an error there, but that doesn't mean I didn't screw something up. Looking around I see "cov.row_names" so I tried "cov.col_names" and voila, it worked.

pyemu/utils/helpers.py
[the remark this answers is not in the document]
briochh: Ah yeah, I see that downstream greostruct_draws does accept tpl-csv pairs. unfortunately the specsim method in this method is expecting par_data_dfs so the tpl-pairs won't work if using specsim... This is another enhancement we could add down the line. For now might be better to have struct_dict here as a dict of lists of dfs (e.g. {GeoStruct: list[par_dfs], ...}) and not the optional use of til-input file pairs, if not using specsim...

pyemu/en.py
briochh: as `Cov` is the expected form `col_names` should be the same as `row_names` and both should be covered by that `names` property. So `col_names` should work but `names` is a handy implicit test for that `Cov` variable actually being a `Cov` object.

briochh (approved):

briochh:
Thanks @wkitlasten, looking good.

PR #517, opened 2024-06-13
fix for pstfrom+usg

jtwhite79:


briochh:
good catch, let's keep an eye on that usg test intermittent fail with xfail off.

Issue #521, opened 2024-06-25
numpy 2.0 incompatibility


andyrich:
# numpy dependency should be set to numpy<2.0

result when importing with numpy=2.0:
![image](https://github.com/pypest/pyemu/assets/16653125/2b21f476-cfa4-4d90-a33f-ecaf3e0fbc08)
![image](https://github.com/pypest/pyemu/assets/16653125/8ca8db7d-f402-4761-bed6-5383901af222)

result when importing with numpy<2.0
![image](https://github.com/pypest/pyemu/assets/16653125/2808624c-2331-4809-9aca-a2282df3cad8)

jtwhite79:
Thanks Andy.  I'm not surprised there are some issues with numpy 2.  We will start working on this but in the meantime we will probably just require numpy<2

PR #526, opened 2024-07-19
Temp pin in numpy to < 2.0

briochh:
belated WIP on some numpy 2+ compat changes


briochh:
temp fix for #521 -- current incompatibility with numpy 2.0 still WIP -- relating to new interaction between new explicit numpy dtypes: https://numpy.org/neps/nep-0051-scalar-representation.html and pandas and our own attempt to support mixed dtypes.

PR #528, opened 2024-07-25
Feat normtransf optional extrapolation

rhugman:
- Added optional extrapolation for normal score transform in `Eds.prep_for_dsi()`. Can be None, linear or quadratic.
- Added optional ZtZ method from pestutils DSI2

### DIFF
--- a/autotest/la_tests.py
+++ b/autotest/la_tests.py
@@ -629,7 +629,7 @@ def ends_freyberg_dsi_test(tmp_path):
     #shutil.copy2(os.path.join(test_d,"pestpp-ies.exe"),os.path.join(t_d,"pestpp-ies.exe"))
 
     pst = pyemu.Pst(os.path.join(t_d,"dsi.pst"))
-    pst.control_data.noptmax = 3
+    pst.control_data.noptmax = 0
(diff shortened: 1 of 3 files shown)


briochh:
@rhugman, is this one gtg? I know that it breaks with our tradition but do you reckon you could add some lines to the `prep_for_dsi()` docstring on the `ztz` and `energy` options and when they might be needed/preferred.

rhugman:
> @rhugman, is this one gtg? I know that it breaks with our tradition but do you reckon you could add some lines to the `prep_for_dsi()` docstring on the `ztz` and `energy` options and when they might be needed/preferred.

That might be helpful hey... :) roger wilco

rhugman:
> @rhugman, is this one gtg? I know that it breaks with our tradition but do you reckon you could add some lines to the `prep_for_dsi()` docstring on the `ztz` and `energy` options and when they might be needed/preferred.

Should be good to go now.

rhugman:
Should be good to go now. Tho it might not be a bad idea if someone double-checked my math...

briochh:
I'm good, if you are good @rhugman!

PR #535, opened 2024-09-12
Feat ppu2

jtwhite79:
Initial support to automatically use ppu for kriging solve if available @briochh it's a hack- open to suggestions


etc/environment.yml
```diff
@@ -4,7 +4,7 @@ channels:
 dependencies:
   # required
   - python>=3.8
-  - numpy
+  - numpy<2.0
```
briochh, line 7: I think, we are supporting numpy 2 now... or have you detected issues?

pyemu/utils/geostats.py
```diff
@@ -872,6 +873,8 @@ def calc_factors_grid(
         var_filename=None,
         forgive=False,
         num_threads=1,
+        try_use_ppu=False,
```
briochh, line 876: let's aim to phase this the default to True, over time.
Any danger of some docstring additions though?

pyemu/utils/geostats.py
```diff
@@ -940,6 +943,58 @@ def calc_factors_grid(
...
+                                                          self.point_data.y.values,zns,
+                                                          x.ravel(),y.ravel(),znt,
+                                                          vartype,krigtype,
+                                                          v.a,v.anisotropy,v.bearing,
+                                                          search_radius,maxpts_interp,
+                                                          minpts_interp,ppu_factor_filename,
+                                                          factorfiletype)
+            plib.free_all_memory()
+            assert os.path.exists(ppu_factor_filename)
+            reformat_factorfile(x.shape[0], x.shape[1], self.point_data, self.geostruct, ppu_factor_filename)
```
briochh, line 992: is the reformatting for downstream compatibility? with our current pyemu methods? pestutils? is there anything we can do about that?

etc/environment.yml
jtwhite79: No issues - this was a legacy thing that snuck in.  We are good with np > 2 yea?

pyemu/utils/geostats.py
jtwhite79: Good points.  I was thinking of a "soft launch" here to ease into it but maybe a docstring would be less ambiguous

pyemu/utils/geostats.py
jtwhite79: I tried to find a way around this but the new fac file format coming out of ppu doesn't have array shape info in the header so I thought it was easiest to patch that here when we know the shape.

etc/environment.yml
briochh: I "_think_" so

jtwhite79:
@briochh the one thing this PR doesn't do (yet) is weave into `PstFrom.add_parameters()`.  I think it could be as simple as passing an explicit `try_use_ppu=True` where we call `calc_factors_grid()` - probably would be wise to make a new optional arg for `add_parameters()` until we get some more in-the-field testing and experience with it (maybe call it `try_use_ppu` also?).  That sound ok with you?

Issue #532, opened 2024-08-22
Idea to move from `os.system` to `sp.Popen` in os_utils.run :)


p-ortega:
Hey Brioch, Just found this behavior and I reckon it could add to the argument to move os system to subprocess. 

Lemme know what do you think 

Cheers

**Description:**
platform: Windows

Running the test from `util_tests.py::test_run`  [`pyemu.os_utils.run(echo test)`] returns `'exe test'` instead of `'test'` without raising an exception. This is coming from the function adding `.exe` to the string if not defined.

I reckon this is where `subprocess` could come in handy. The same function with `subprocess.Popen` returns `'file not found'`.

briochh:
So, I guess the subprocess option would remove the need to append that ".exe" string on windows? But it does require careful splitting of call strings to lists of args, right? Is that generalisable across use-cases -- with all possible exe and option combos?
Not withstanding the potential can or worms, I am keen to see where such a shift might take us. 

One thing to watch out for is maintaining how we deal with launching runs inside another directory and the logic behind where executables are on the PATH, from that other dir. - this has been a pain point for me when using the `sim.run()` methods in flopy.

jtwhite79:
was thinking about this - what if for now we have an optional kwarg that gets passed to `pyemu.os_utils.run()` like "use_sp" so that the `os.system` and `sp.Popen` styles can co-exist? This would give us time to take it out in the wild and try it on a range of cases.  thoughts?

p-ortega:
Sounds like a solid approach -- if ain't broke! 

I’m up for giving it a shot and seeing how it works. I can try it out on a model I’m re-HM’ing soon and see how it goes

PR #539, opened 2024-10-07
Feat run_sp

p-ortega:
**This PR adds the `os_utils.run_sp()` function to run commands using `subprocess.Popen` as discussed in [532](https://github.com/pypest/pyemu/issues/532). The main function (`os_utils.run()`) will still use `os.system` by default, or call `run_sp()` if the `run_sp` option is set to `True`.**

Four unit tests were added to check stdout, log file output, and assert run success.

Additionally, `run_sp()` has been tested with some GMDSI tutorials, but I still need to test it on a real model.

Let me know what you think!

### DIFF
--- a/autotest/utils_tests.py
+++ b/autotest/utils_tests.py
@@ -2054,20 +2054,32 @@ def run_test():
     else:
         raise Exception("should have failed")
 
-def run_sp_test():
-    import pyemu
+def run_sp_success_test():
     import platform
-
     if "window" in platform.platform().lower():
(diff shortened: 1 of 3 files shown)


p-ortega:
I need to fix my tests for macos and ubuntu .. I'll be back

jtwhite79:
All good P!  Thanks for working on this!

jtwhite79:
@p-ortega is the good to go?

p-ortega:
@jtwhite79 I think is g2g now

Issue #315, opened 2021-12-30
Use `logging` from standard library


Hugovdberg:
`PyEmu` implements a custom logger to log messages from various parts of the package. Python comes with a `logging` module that provides the same functionality, and more. Also, `PstFrom` is currently hardcoded to log to a file in the current working directory, as well as an optional logging to `stdout`. I'm using `pyemu` within a `snakemake` workflow, with log files being stored in a dedicated directory, so I would like to be able to specify how and where the messages are logged. The standard library also provides means to configure different log levels for different streams (for example debug and info messages only to file, but warnings and errors also to the screen, or even custom log handlers).

Would you guys be open to a pull request that replaces the current logger with the built-in `logging` interface? It would mean that some options need to be removed, such as the `echo` argument to various helpers. We would also need to add a small tutorial for people who would like to keep the current functionality of logging to both a file and stdout, but in return the users get more flexibility both in what is logged as well as how it is logged. This removes some (in my humble opinion ;-) ) unneccessarily opnionated implementation from library code, which makes the library more pythonic.

jtwhite79:
sounds good to me as long as we can still get timestamps and weave the logger into the helpers?  Also, we need to retain the snarky and insulting messages! haha

@briochh?

Hugovdberg:
The nice thing about the standard logging module is that it allows you much more fine grained control over the log messages. You can keep logging all messages, from anywhere in the library, but the user of the library gets more control how and where the messages are going. 

It is not necessary for the library to provide timestamps as they can be included by the handler. The default line format does not include the timestamp, but it is easily added. It might be useful to provide a helper function that sets up the logging module similar to the current format so those who want the `pyemu` recommended settings can simply call `pyemu.set_default_logging`, and those who want more control setup `logging` as they wish.

I'll see if I can whip up a draft pull request for the `pst_from` module soon, so you can see how it would work.

Hugovdberg:
@jtwhite79 what versions of Python should be supported? The travis config suggests Python versions 2.7, 3.5, 3.6 and 3.7 are used for testing, but the appveyor config suggests only Python 3.7 is tested. Also, `PstFrom._setup_dirs` uses f-strings, so that will not run pre-Python 3.6. Can I assume 3.7+?

jtwhite79:
I think 3.7+ is fine.  We are really only using github actions anymore....

briochh:
Sounds like a solid plan to me @Hugovdberg. Thanks for having a dig into this.

PR #544, opened 2024-10-12
First try at changing the logging facilities

Hugovdberg:
Ages ago I proposed to migrate to the built in logging facilities to allow for more flexible logging (see #315). Recently I have been using `pyemu` again and faced the same issue I faced then so here is a PR (work in progress) to add logging instead of writing files directly. 
I decided to first tackle the new `run_sp` method as this was the newest so least likely to break peoples workflows should it be not as stable as I think it is (see #539 / #532).

### DIFF
--- a/pyemu/utils/log_utils.py
+++ b/pyemu/utils/log_utils.py
@@ -1,0 +1,37 @@
+"""Utilities to help with logging."""
+
+import logging
+import sys
+from typing import Optional, Union
+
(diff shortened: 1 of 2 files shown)


p-ortega:
Hey @Hugovdberg, I think you would have to modify or remove this test, pytest::run_sp_capture_output_test, for the CI workflow to work with the new logging capabilities.

Hugovdberg:
Probably a lot of tests will fail once we add this in other places as well :-) testing output is always troublesome..

PR #546, opened 2024-10-18
Feat ppu3

jtwhite79:
More direct and involved coupling between pstfrom and ppu pilot points.  Very rough, looking for help/suggestions...


autotest/utils_tests.py
```diff
@@ -2570,8 +2570,73 @@ def thresh_pars_test():
...
+def ppu_geostats_test(tmp_path):
+    import sys
+    import os
+    import numpy as np
+    import matplotlib.pyplot as plt
+    import pyemu
+    
+    import flopy
+
+    sys.path.insert(0,os.path.join("..","..","pypestutils"))
```
briochh, line 2587: Note for self: 
I think, shouldn't need this in the CI now -- we `pip install -e` the pypestutils directory directly so `import pypestutils as ppu` should "work". 

For a local install of ppu: `pip install -e {path to pypestutils setup.py}`

pyemu/utils/pst_from.py
```diff
@@ -1830,6 +1839,10 @@ def add_parameters(
         comment_char=None,
         par_style="multiplier",
         initial_value=None,
+        prep_pp_hyperpars=False,
+        pp_options={},
```
briochh, line 1843: Note to self:
Would be keen to support expression of `pp_space` and `use_pp_zones` in `pp_options` dict. Just need to make sure backcompat and decide on which takes priority if both are present.

autotest/utils_tests.py
jtwhite79: I agree. That was just for me locally while we were waiting to a new ppu version to be released.  It's in several fxns so let's strip those out once things stabilize a bit more.

pyemu/utils/pst_from.py
jtwhite79: Good call.  Maybe we can mark pp_space and use_pp_zones as deprecated and just map their values in pp_options for now?

briochh: yeah, I think that should work. We should make that we document all the possible pp keywords but suspect this will be a moving target for a bit.

Issue #527, opened 2024-07-25
Add python function (to PstFrom.add_py_function) as a function/string. 


martinvonk:
Currently, the `PstFrom` method  `add_py_function` takes a path to a separate python source file. This python source file is read, interpreted and copied to the forward run script.  

I think it would be nice to allow for adding the code from the python source file as a string to the PstFrom class. For instance, say you have a file `_update_model.py` containing the following code:

```python
def run():
    updated_parameters = read_csv("parameters.csv")
    new_simulation = some_model.simulate(parameters=updated_parameters)
    new_simulation.to_csv("simulation.csv")
```
This file is added to the PstFrom class via:
```python
pf = PstFrom()
pf.add_py_function("_update_model.py", "run()", is_pre_cmd=None)
pf.mod_py_cmds.append("run()")
```

My proposal would be to allow to create a new method add_py_function_str:
```python
update_model_str = """
def run():
    updated_parameters = read_csv("parameters.csv")
    new_simulation = some_model.simulate(parameters=updated_parameters)
    new_simulation.to_csv("simulation.csv")
"""
pf = PstFrom()
pf.add_py_function_str(update_model_str , "run()", is_pre_cmd=None)
pf.mod_py_cmds.append("run()")
```
This way, you don't need a separate python source file per se. You only need one python script / notebook which makes examples and changing the code easier. 

Also, to implement it, the code does not need too many changes. Only the [interpretation of the file (or string)](https://github.com/pypest/pyemu/blob/5551e161bb4978c599b7352ee373b0d88d831d68/pyemu/utils/pst_from.py#L1267-L1293) has to be a separate function which is called by both `add_py_function` and `add_py_function_str`.

martinvonk:
I figure I could also just write the string to a file and parse that to the `add_py_function` method. So if that is a better convention, this issue can be closed.

jtwhite79:
I see what you are getting @martinvonk .  From my standpoint, I always test run the function during the setup to make sure its doing what its supposed to do - that's really why we look for an existing function.  Note that function can be in the py script that is calling pstfrom (one in the same script) but in the case of a notebook, I can see the value in not forcing the function to be in a py file.. @briochh any thoughts on this?  I reckon we could make this a user beware thing if the first arg to `add_py_function()` gives file not found...

martinvonk:
[...]

PR #549, opened 2024-11-14
allow parsing function to add_py_function

martinvonk:
This PR allows for parsing a function/callable to the `PstFrom.add_py_function` method. 

Originally, only a file name is allowed. This (separate) file was read, and the corresponding function in the file was added to the PstFrom class. Now you can directly add a function, available in the same script as the one where the PstFrom class is initialized. 

Solves #527


briochh:
Hi @martinvonk, this looks like a nice touch. Do you think you could add to or amend a test to run this option? There should already be one in pst_from_tests that is hackily doing it the old way with a function in the current file.

briochh:
Thank @martinvonk, good addition!

PR #559, opened 2024-12-06
Feat dsi_pyworker

rhugman:
- relies on JDub's PR #557 
- added dsi_pyworker() to helpers
- switched dsi tests in la_tests to use the pyworker
- painted everything with red stripes for extra speeeeddddd


briochh:
thanks @rhugman, getting the occasional fail in the dsi tests, unsure where they come from. e.g. https://github.com/pypest/pyemu/actions/runs/12248019374/job/34166881016#step:14:3319 
Looks like it is triggered by a failure during the dsi runs. Maybe overlapping file access. Maybe memory issue when running with a few workers. Hopefully using pypestworker approach here heads it off. Let's keep an eye on it though.

PR #560, opened 2024-12-06
Appended some small changes to account for 1d array shapes from unstructured grids

jonathanqv:


briochh:
Thanks @jonathanqv , good to have your eyes on the unstructured stuff!

PR #561, opened 2024-12-11
removing rc_params setting in plotutils

### LINKED ISSUE #554: Please don't modify the matplotlib preferance settings before
Importing pyemu changes the matplotlib font size which overides user preferences and can cause inconsistancy in the the users figures... it's really annoying.  **Please don't do this.**  If you need it for examples etc., please make sure the changes are within the `if __name__ == '__main__':`

File: 
..pyemu/plot/plot_utils.py

Lines 12-18

```
font = {"size": 6}
try:
    import matplotlib

    matplotlib.rc("font", **font)

    import matplotlib.pyplot as plt
    from matplotlib.backends.backend_pdf import PdfPages
    from matplotlib.gridspec import GridSpec
except Exception as e:
    # raise Exception("error importing matplotlib: {0}".format(str(e)))
    warnings.warn("error importing matplotlib: {0}".format(str(e)), PyemuWarning)
```

briochh:
Moved Matplotlib font.size rc_param setting to a decorator and wrapper to run each plot_utils function in a rc_context. The net effect on plots should be the same. The decorator can take a dict of pars that would ordinarily get passed through to `rc_context()` if we want function-wide control at any point. This should close #554


hansonmcoombs:
Thanks heaps!

Issue #564, opened 2024-12-16
mfile_skip getting ignored in pf.add_parameters()


kmarkovich:
Hi, I am trying to add parameters from an ascii file (attached) that has some rows at the top that need to be skipped. Getting an error "could not convert string to float: 'ncols'" which suggests that pyemu is still trying to read that first line. 
[AWC_subset.txt](https://github.com/user-attachments/files/18151597/AWC_subset.txt)

briochh:
hey @kmarkovich, 
Thanks! looks like you have caught one! I'll get a fix plumbed in.

kmarkovich:
right on, thanks @briochh !

kmarkovich:
@briochh mfile_skip should still preserve those skipped lines when writing out the file, no? looks like it's dropping them in the written out file

briochh:
it's a good point @kmarkovich, I think the answer should be probably, yes. Seems our (lazy) preservation of headers might have been limited to list-like files. I'll look at a patch. Might take a moment.

briochh:
@kmarkovich, I have a probable fix pushed to develop now. See how this works for you.

kmarkovich:
works like a dream!

PR #566, opened 2024-12-18
Propagating headers in array files to model input files

briochh:
+ bit of work around tpl files including header for direct array style pars
+ added too test to include direct add_pars and pp add_pars


briochh:
hopefully further addresses #564

PR #567, opened 2024-12-19
Update to spatial ref arg doc string 

briochh:
This was a long-orphaned commit sitting on my fork, just cleaning up. -- still appropriate?


jtwhite79:
looks good to me!

PR #598, opened 2025-06-06
small fixes to pyemu.helpers.prep_for_gpr(), gpr CI tests, and gpr hosaki notebook. 

kmarkovich:
prep_for_gpr now saves out a generic obj name to pickle files to avoid char length limits and has an added feature to copy over the pestpp bins from the template dir, because prep_for_gpr() was building a fresh template with no binaries in it and failing the noptmax 0 test. updated gpr tests and hosaki NB to include new t_d arg, added option in hosaki NB to grab the pestpp bins if not in path

### DIFF
--- a/autotest/utils_tests.py
+++ b/autotest/utils_tests.py
@@ -2812,7 +2812,7 @@ def gpr_compare_invest():
 
     pst_fname = os.path.join(m_d,case+".pst")
     gpr_t_d = os.path.join(case+"_gpr_template")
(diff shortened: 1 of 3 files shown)


briochh:
@kmarkovich, I think the failing tests are unrelated. I have just submitted a PR to try and fix...then we can get this merged in.

briochh:
Thanks @kmarkovich

PR #606, opened 2025-07-11
phi components to include prior info if pestmode is "reg"

briochh:
I think we wan't to include prior information if in reg mode right?


briochh:
scratch that

PR #607, opened 2025-07-30
Feat parobs

jtwhite79:
added Pst fxn to add par values as obs to help with bound-penalties, a lil sugar in result handler, some other minor things


briochh (approved):
minor worry about pytest creep but let's tackle that later

Issue #605, opened 2025-07-08
pestpp_options: spaces unhandled


cnicol-gwlogic:
Given this example pestpp control file version 2 entry:
```opt_dec_var_groups             dv_drcq, dv_irrig``` 
Noting the space after the comma in the comma-separated list of variables, Pyemu 1.3.8 only parses out the first option value in the list. If the space is removed after the comma, the entire list is parsed ok.

p-ortega:
@briochh  Do you want me to have a look at this? I have a bit of time atm

briochh:
Sorry @cnicol-gwlogic, I missed this. @p-ortega go for it!

cnicol-gwlogic:
No stress, minor issue, easily fixed by me not putting spaces where they don't belong!

PR #608, opened 2025-08-01
Improve parsing of pestpp_options with spaces for pst v2

p-ortega:
### Problem
Previously, `pyemu`'s handling of `pestpp_options` in `.pst` files (version 2) only supported comma-separated values **without spaces**. For example:

    lambdas                       1,2,10,999
    opt_dec_var_groups           dv_drcq,dv_irrig

If values included **spaces after commas**, like:

    lambdas                       1, 2, 10, 999
    opt_dec_var_groups           dv_drcq,   dv_irrig

Then only the first token (before the space) was parsed, the rest were ignored.

---

###  What’s changed

Now the parser accepts `pestpp_options` with optional spaces after commas. For example, the following now parse correctly:

    lambdas                       1, 2,10,999
    opt_dec_var_groups           dv_drcq,   dv_irrig
    panther_transfer_on_finish   file1, file2, file3
    ies_n_iter_reinflate         3, 5,  999
    ies_reinflate_factor         0.1 , 0.05, 0.1

---

### Tests

- Added one unit test to ensure parsing works with and without spaces.
- Verified that existing behavior is unchanged (passing full stack for now).

This would `close` [#605](https://github.com/pypest/pyemu/issues/605)


p-ortega:
@briochh Not sure if pestpp actually accepts pestpp_options that random

pyemu/pst/pst_controldata.py
```diff
@@ -10,6 +10,7 @@
 import warnings
 import numpy as np
 import pandas
+import re
```
briochh, line 13: is this import required?

autotest/pst_tests.py
```diff
@@ -819,6 +819,25 @@ def new_format2_test(tmp_path):
         raise Exception("fail:"+str(e))
     os.chdir(b_d)
 
+def new_format_pestpp_options_test(tmp_path):
```
briochh, line 822: Don't think `tmp_path` is needed in this instance. This is a pytest thing and is useful if modifying files, to prevent test interference. Here just reading the reference file so should be safe without (providing that files isn't getting beaten up by another test).

briochh (commented):
@p-ortega, yeah may not be "real" pestpp args but I don't really think we are doing much control file checking in general in pyemu...

pyemu/pst/pst_controldata.py
p-ortega: good catch .. lazoo coding

PR #626, opened 2025-10-22
Feat easiertesting

rhugman:
Because I keep forgetting how to make local tests work.

 - script setups up pyemu mamba/conda env
 - adds pyemu bin dir to the env PATH
 - calls get_pestpp to get pestpp binaries

No longer need to set `local_tests==True`.


rhugman:
> Sorry @rhugman, a little slow getting to this. I have a PR at #619 around the testing and exe locations (I suspect with a similar motivation). Likely that the two PRs will be in conflict, do you think you could merge my branch into your PR? Get rid of anything that doesn't work for you. In the longer term, I have been playing with uv a little; looks like it might provide a good solution for rapidly setting up test envs.

merged ur branch. I dont think they conflict. Different solutions to the same problem. It may be a matter of choosing whether my implementation fits. I essentially added a helper script to add the repo bin dir to the mamba env PATH + get_pestpp.

Also - seems someone changed python version reqs for pypestutils. 3.9 no longer supported?

PR #641, opened 2025-11-29
Tensor interpolation for non-stationary geostat fields

wkitlasten:
Some utilities to perform log-Euclidean interpolation of anisotropic variance tensors from conceptual points to a regular grid. Also performs spatial convolution to create non-stationary geostatistical fields from variance tensors. Uses pypestutils where possible, with fallback to python implementation of moving average if needed (e.g., pypestutils fieldgen2d_sva does not accept external iid fields).

### DIFF
--- a/pyemu/utils/nsaf_helpers.py
+++ b/pyemu/utils/nsaf_helpers.py
@@ -14,7 +14,8 @@ from nsaf_utils import (
 )
 
 
-def generate_fields_from_files(tmp_model_ws, model_name, conceptual_points_file,
+def generate_fields_from_files(model_name, conceptual_points_file,
+                               model_ws='.',
                                zone_files=None, iid_files=None, field_name=['field'],
                                layer_mode=True, save_dir=None, tensor_interp='krig',
                                transform='log', n_realizations=1, ml=None, sr=None,
@@ -26,7 +27,7 @@ def generate_fields_from_files(tmp_model_ws, model_name, conceptual_points_file,
 
     Parameters
     ----------
-    tmp_model_ws : str
+    model_ws : str
         Path to model workspace containing flopy model files
     model_name : str
         Name of the flopy model (e.g., 'freyberg6')
@@ -52,7 +53,7 @@ def generate_fields_from_files(tmp_model_ws, model_name, conceptual_points_file,
     layer_mode : bool, default True
         If True, process each layer independently
     save_dir : str, optional
-        Directory path to save output files (default: tmp_model_ws)
+        Directory path to save output files (default: model_ws)
     tensor_interp : str, default 'krig'
         Tensor interpolation method
     transform : str, default 'log'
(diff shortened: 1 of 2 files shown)


pyemu/utils/pst_from.py
```diff
@@ -2605,7 +2605,7 @@ def add_parameters(
                     #if "pypestutils" not in self.extra_py_imports:
                     #    self.extra_py_imports.append("pypestutils")
                     print(config_df_filename)
-                    config_func_str = "pyemu.utils.pp_utils.apply_ppu_hyperpars('{0}')".\
+                    config_func_str = "pyemu.utils.pp_utils.apply_ppu_hyperpars_tensor('{0}')".\
```
briochh, line 2608: breaks current non-stationary hyper par implementation... I think best to option "tensor" (at least until it is proven)

pyemu/utils/nsaf_helpers.py
briochh, line 1: wondering if these nsaf funcs would fit better within pypestutils or a stand alone library (or submodule here). For now might be good to move into a prototypes submodule here. That way you we can keep reasonably contained (incl. dependancies) while refined and testing funcs better targeted.

pyemu/utils/nsaf_utils.py
```diff
@@ -0,0 +1,1679 @@
+import numpy as np
+import pandas as pd
+from scipy.linalg import logm, expm
+import os
+import matplotlib as mpl
+
+mpl.use('Tkagg')
```
briochh, line 7: this will cause failure on some CI platform (on import)

examples/pstfrom_mf6_ppu_tensor.ipynb
```diff
@@ -0,0 +1,2460 @@
...
+     ]
+    }
+   ],
+   "source": [
+    "import os\n",
+    "import shutil\n",
+    "import numpy as np\n",
+    "import pandas as pd\n",
+    "import matplotlib\n",
+    "matplotlib.use('Qt5Agg')\n",
```
briochh, line 29: CI might not like this line

examples/pstfrom_mf6_ppu_tensor.ipynb
```diff
@@ -0,0 +1,2460 @@
...
+    }
+   ],
+   "source": [
+    "tmp_model_ws = \"temp_pst_from_nsaf\"\n",
+    "if os.path.exists(tmp_model_ws):\n",
+    "    shutil.rmtree(tmp_model_ws)\n",
+    "shutil.copytree(org_model_ws,tmp_model_ws)\n",
+    "# copy pestpp and mf6\n",
+    "shutil.copy(r'C:\\bin\\modflow\\mf6.6.2_win64\\bin\\mf6.exe', os.path.join(tmp_model_ws,'mf6.exe'))\n",
+    "shutil.copy(r'C:\\bin\\pestpp\\pestpp-ies.exe', os.path.join(tmp_model_ws,'pestpp-ies.exe'))\n",
```
briochh, line 404: Will need to be smarter here with bin/exe definition for x-platform and CI

examples/pstfrom_mf6_ppu_tensor.ipynb
briochh, line 1: much safer to clean notebook outputs before adding to git!

wkitlasten:
Thanks for taking a look. I'll move it elsewhere.

PR #643, opened 2025-12-08
Better test skipping and basic project linting

mwtoews:
The primary change in this PR is to do better test skipping if optional dependencies or executables are not available.

It also applies a few basic lint fixes discovered (and sometimes auto-fixed) with `ruff check`. So only a small subset of these are addressed in this PR:
```
$ uvx ruff check --statistics
324     F841    [-] unused-variable
239	F401	[-] unused-import
168	E722	[ ] bare-except
 27	E402	[ ] module-import-not-at-top-of-file
 21	E721	[ ] type-comparison
 17	E712	[ ] true-false-comparison
 17	F821	[ ] undefined-name
 16	F523	[-] string-dot-format-extra-positional-arguments
 16	F811	[-] redefined-while-unused
 15	E731	[ ] lambda-assignment
 14	F541	[*] f-string-missing-placeholders
 10	E741	[ ] ambiguous-variable-name
 10	F403	[ ] undefined-local-with-import-star
  5	E713	[*] not-in-test
  4	E711	[ ] none-comparison
  3	    	[ ] invalid-syntax
  2	E401	[*] multiple-imports-on-one-line
  1	E703	[*] useless-semicolon
  1	F402	[ ] import-shadowed-by-loop-var
  1	F602	[ ] multi-value-repeated-key-variable
Found 911 errors.
[*] 305 fixable with the `--fix` option (317 hidden fixes can be enabled with the `--unsafe-fixes` option).
```
where after this PR the remaining are:
```
$ uvx ruff check --statistics
301     F841    [ ] unused-variable
209	F401	[-] unused-import
129	E722	[ ] bare-except
 27	E402	[ ] module-import-not-at-top-of-file
 21	E721	[ ] type-comparison
 17	E712	[ ] true-false-comparison
 15	E731	[ ] lambda-assignment
 14	F821	[ ] undefined-name
 10	E741	[ ] ambiguous-variable-name
 10	F403	[ ] undefined-local-with-import-star
  8	F811	[ ] redefined-while-unused
  1	F602	[ ] multi-value-repeated-key-variable
Found 762 errors.
[*] 187 fixable with the `--fix` option (333 hidden fixes can be enabled with the `--unsafe-fixes` option).
```
(And I don't think all of these need to be addressed and can be ignored).

### DIFF
--- a/autotest/autotest_notebooks.py
+++ b/autotest/autotest_notebooks.py
@@ -23,7 +23,7 @@ os.mkdir(testdir)
 
 
 def get_notebooks():
-    return [f for f in os.listdir(nbdir) if f.endswith('.ipynb') and not "notest" in f]
+    return [f for f in os.listdir(nbdir) if f.endswith('.ipynb') and "notest" not in f]
 
 def run_notebook(fn):
     #pth = os.path.join(nbdir, fn)
--- a/autotest/emulator_tests.py
+++ b/autotest/emulator_tests.py
@@ -15,6 +15,8 @@ ies_exe_path = get_exe_path("pestpp-ies")
 mou_exe_path = get_exe_path("pestpp-mou")
 
 def dsi_freyberg(tmp_d,transforms=None,tag=""):
+    if not ies_exe_path:
+        pytest.skip("missing ies_exe_path")
 
     test_d = "ends_master"
     test_d = setup_tmp(test_d, tmp_d)
@@ -92,6 +94,8 @@ def test_dsi_mixed(tmp_path):
 
 # @pytest.mark.timeout(method="thread", timeout=1000)
 def test_dsivc(tmp_path):
+    if not mou_exe_path:
+        pytest.skip("missing mou_exe_path")
     # basic quick as so can re-run here
     dsi_freyberg(tmp_path, transforms=None)
     # now test dsicv
@@ -205,6 +209,7 @@ def plot_freyberg_dsi():
 
 
 def lpfa_freyberg(tmp_d="temp",transforms=None):
+    pytest.importorskip("sklearn")
 
     test_d = "ends_master"
     test_d = setup_tmp(test_d, tmp_d)
@@ -331,6 +336,8 @@ def test_lpfa_std(tmp_path):
 
 
 def gpr_compare_invest():
+    if not mou_exe_path:
+        pytest.skip("missing mou_exe_path")
     import numpy as np
     from sklearn.gaussian_process import GaussianProcessRegressor
     case = "zdt1"
@@ -492,6 +499,8 @@ def gpr_compare_invest():
 
 
 def gpr_constr_invest():
+    if not mou_exe_path:
+        pytest.skip("missing mou_exe_path")
     import numpy as np
     from sklearn.gaussian_process import GaussianProcessRegressor
     case = "constr"
@@ -812,14 +821,13 @@ def gpr_zdt1_test():
     gpr_t_d = os.path.join(case + "_gpr_template")
 
     data, input_names, output_names = collate_training_data(pst,m_d,case)
-    from pyemu.emulators.gpr import GPR
     gpr = GPR(data=data.copy(),
           input_names=input_names,
           output_names=output_names,
           #transforms=transforms,
           #kernel=gp_kernel,
           n_restarts_optimizer=20,
-          );
+          )
     gpr.fit()
     gpr.prepare_pestpp(m_d,case,gpr_t_d=gpr_t_d)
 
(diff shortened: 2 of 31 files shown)


briochh:
thanks for this @mwtoews. The codebase could defo benefit from a clean up, I am wondering how tolerant we want to be of failed imports and missing binaries though? Those red 'x's are a pretty strong motivator to fix broken links between packages and dependancies.

autotest/autotest_notebooks.py
briochh, line 1: I think this notebook wrapper is no longer hit -- at some point, we should check and remove,

autotest/emulator_tests.py
```diff
@@ -15,6 +15,8 @@
 mou_exe_path = get_exe_path("pestpp-mou")
 
 def dsi_freyberg(tmp_d,transforms=None,tag=""):
+    if not ies_exe_path:
+        pytest.skip("missing ies_exe_path")
```
briochh, line 19: assume we will throw a fail elsewhere in the CI if bins are missing -- might be worth checking this.

autotest/emulator_tests.py
```diff
@@ -331,6 +336,8 @@ def test_lpfa_std(tmp_path):
 
 
 def gpr_compare_invest():
+    if not mou_exe_path:
+        pytest.skip("missing mou_exe_path")
```
briochh, line 340: These "invest" funcs shouldn't be collected by pytest in the first place, so the "skip" probs ineffective.

autotest/full_meal_deal_tests_ignore.py
briochh, line 1: I think this is also mostly legacy func testing -- at some point we should check and remove

autotest/pst_from_tests.py
```diff
@@ -607,7 +600,7 @@ def add_py_function_test(tmp_path):
 def test_mf6_freyberg(tmp_path):
     # import sys
     # sys.path.insert(0, os.path.join("..", "..", "pypestutils"))
-    import pypestutils as ppu
+    ppu = pytest.importorskip("pypestutils")
```
briochh, line 603: we probably don't want to be skipping this test if import fails. 

Currently, there is the pyemu geostats interp fallback if ppu is not available (though this may be deprecated in the future if we want to lean more on ppu)

So here, I think we either want to raise if import fails or continue (which would mean that we are testing the pyemu fallback).

autotest/pst_from_tests.py
```diff
@@ -1279,15 +1272,12 @@ def mf6_freyberg_shortnames_test(tmp_path):
 
 
 def mf6_freyberg_da_test(tmp_path):
+    flopy = pytest.importorskip("flopy")
```
briochh, line 1275: I think for these we can just do a straight import flopy. flopy should be defined as a testing dependency (only) so if you are running these tests you do need flopy so these would be better to raise rather than skip?

autotest/pst_from_tests.py
```diff
@@ -3526,19 +3491,16 @@ def mf6_freyberg_arr_obs_and_headerless_test(tmp_path):
...
     pd.set_option('display.width', 1000)
-    #try:
-    import flopy
-    #except:
-    #    return
 
     # import sys
     # sys.path.insert(0,os.path.join("..","..","pypestutils"))
-    import pypestutils as ppu
+    ppu = pytest.importorskip("pypestutils")
```
briochh, line 3503: ditto above, either we should be raising here if ppu absent or (for now) continuing and allowing pyemu geostationary fallback

autotest/pst_from_tests.py
briochh, line 1: generally for the tests in here we rely on flopy (which should be a testing-only dependency for pyemu). So I think ok to raise rather than skip if flopy import fails. For dealing with ppu import fails at the moment (while we still have the pyemu geostat fallback) I think we can continue but in the future we could look to retire the pyemu fallback and require ppu [perhaps we can set up a discussion on this?], in which case raising on ppu import fails would be appropriate.

autotest/utils_tests.py
briochh, line 1: same as in pst_from_tests regarding flopy and ppu imports -- maybe raising is preferable to skipping?

briochh (commented):
@mwtoews, thanks for these improves. See comments around flopy and ppu import fail skips... think we might be better off with a non-skipping option in most cases?

briochh:
@mwtoews, a lot has changed in on dev branch since this PR. Still some valuable stuff in here though. Worth rehashing?

mwtoews:
I'll revise this PR when I get a chance, not this week though.

PR #649, opened 2025-12-12
Declare `scikit-learn` as an optional dependency

nathanjmcdougall:
`scikit-learn` is used in the `pyemu.emulators` module, among other places:

https://github.com/pypest/pyemu/blob/5dd3c785991ec991974c53cd0b1058281a2f872f/pyemu/emulators/__init__.py#L47-L55

`graphviz` is just used in `misc\graph_pyemu.py`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,6 +38,7 @@ optional = [
     "jinja2",  # for to_latex options
     "pypestutils",
     "scikit-learn",
+    "graphviz",
 ]
 test = [
     "coveralls",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,7 +36,8 @@ optional = [
     "scipy",
     "shapely",
     "jinja2",  # for to_latex options
-    "pypestutils"
+    "pypestutils",
+    "scikit-learn",
 ]
 test = [
     "coveralls",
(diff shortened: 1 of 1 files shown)


briochh:
my feeling is that the graph_pyemu isn't picked up anywhere. Let's check in with @aymanalz, we maybe be able to utilise it or drop graphviz as a dependancy.

nathanjmcdougall:
Yeah I get the impression that the module there is intended for as a dev tool - to visualize the architecture of the modules.

I think it would be helpful to adopt `dependency-groups` from [PEP 735](https://peps.python.org/pep-0735/#example-dependency-groups-table) to give dedicated declaration sections for dev dependencies that aren't relevant for consumers of pyEMU as a package, but only for developers of pyEMU itself. Similarly, the test dependencies would probably be better as in a  `dependency-groups` group than as optional dependencies.

This would allow us to declare lint dependencies etc. (e.g. if Ruff gets added) and so `graphviz` could be declared without adding it to the actual dependencies of the package. I'd put in a separate PR for that if you want it.

briochh:
I have a fork, that is slowly (and tentatively) exploring a move to uv for the whole project, and thus an updated (and more responsible pyproject.toml) https://github.com/briochh/pyemu/tree/feat_uv . perhaps we can bite the bullet on that at some point.

nathanjmcdougall:
That would be excellent. Probably in that case, we can hold off on declaring `graphviz` for now. I'll revert it when I get a moment.

aymanalz:
Thank you for checking about Graphviz dependency. I do not think I will use this package.

PR #653, opened 2025-12-12
Improve developer experience for running binaries from test suite

nathanjmcdougall:
I've added a bit of documentation about test dependencies to the README, and added a `.gitignore` entry for pypestworker log files that get generated.

Although perhaps, rather than adding `autotest/pypestworker_*.txt` to `.gitignore`, it would be better to reconfigure the workers to write the log files to `autotest/temp` - happy to take advice on that.

I tried building `MODFLOW-USG` from source without much luck - so I tried `git checkout 96b16ef1544277a5cf64642b3aa5b5d2f89c4009 -- bin/win/mfusg.exe` to get an old binary and renamed it to `mfusg_gsi.exe` ... happily it worked and I was able to pass `autotest\pst_from_tests.py::usg_freyberg_test` :)
It would be good if you could provide advice on this to improve/correct the information I've got in the README for this one.

Finally, there's a mention of `mt3dusgs` here:
https://github.com/pypest/pyemu/blob/5dd3c785991ec991974c53cd0b1058281a2f872f/autotest/conftest.py#L63-L72

But I can't find `mt3dusgs` referenced anywhere else - maybe it can be removed?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -66,6 +66,7 @@ pyemu/.idea
 
 autotest/*.bin
 autotest/*.dat
+autotest/pypestworker_*.txt
 autotest/temp
 autotest/.noseids
 autotest/temp*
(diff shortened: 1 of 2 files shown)


briochh:
thanks @nathanjmcdougall, 

pypestworker log files are default written to current directory so I think ok to gitignore for local tests.

for the modflow bins, the CI leans on https://github.com/modflowpy/install-modflow-action. For local tests maybe we should be pointing developers to the https://github.com/MODFLOW-ORG/executables/releases source. There is also a flopy get-modflow utility that users could leverage.

probably the mt3dusgs reference could be dropped -- I think the test suite is covered off by mf6 and some mfnwt

PR #657, opened 2025-12-15
Update description of current distributions included from `usgs/pestpp`

nathanjmcdougall:
When using `pestpp`, the default behaviour doesn't seem to include any executables except those from `pestpp`, e.g. it doesn't include `mf6`. I'm guessing that these executables are no longer distributed in the releases in the upstream `usgs/pestpp` repo, and that these docs are just a bit out of date - so trying to update them.

My guess might be wrong - if there's a way to download other relevant executables using `get-pestpp` it would be helpful to incorporate into #653

### DIFF
--- a/get_pestpp.md
+++ b/get_pestpp.md
@@ -73,7 +73,7 @@ Other auto-select options are only available if the current user can write files
 
 ## Selecting a distribution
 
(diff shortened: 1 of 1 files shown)


briochh:
good catch, yeah get-pestpp only deals in pestpp binaries. modflow binaries can be installed from https://github.com/MODFLOW-ORG/executables or using https://flopy.readthedocs.io/en/latest/md/get_modflow.html. 
from a dependency point of view pyemu is more tightly linked to PEST(++) than to modflow. Modflow is just our most widely used forward simulator and therefore the one we engage with in the tests.

PR #660, opened 2025-12-15
Feat frunphi

jtwhite79:
added super hack to get weighted phi as an observation.  refactored draw_new to be at the `Ensemble` level with derived methods for `ParameterEnsemble` and `ObservationEnsemble`.  Some small mods to `pars_as_obs()` so that user-tag can be added to the resulting obs names.


pyemu/en.py
```diff
@@ -833,6 +909,34 @@ def from_gaussian_draw(
...
+        in `self`.  
+
+        Args:
+            num_reals (int) : number of realizations to generate
+            include_noise (varies): a bool or a float the describes the standard deviation of
+                noise to add to the new realizations.  This is to help with the issue of 
+                under-varied new realizations resulting from npar >> nreals in `self`. If True,
+                The standard devation is set to one over the square root on number of reals in 
+                `self`.  
+            noise_reals (ParameterEnsemble): other existing realizations (likely prior realizations)
```
briochh: should this be ObservationEnsemble now?

pyemu/en.py
```diff
@@ -833,6 +909,34 @@ def from_gaussian_draw(
...
+                that are used as noise realizations in place of IID noise that is used if `include_noise` 
+                is True and `noise_reals` is None.
+        
+        Returns
+            ParameterEnsemble
```
briochh: ObservationEnsemble?

PR #661, opened 2025-12-15
Trying to get some easy wins on the test suite speed

briochh:
Reduced scale of some test and improved some memory and file handling.

Not a huge gain -- just 5 mins or so. 

May mess with draw_new test changes in #660, can merge in later if needed


briochh:
not much to be had here, the big blocker seems to be py 3.12 (especially on win). PR of value though.

PR #664, opened 2025-12-16
Migrate from Sphinx to MkDocs

### LINKED ISSUE #519: Documentation not working
Hi, the link to the documentation seems to be broken, the pdf has a couple of titles but is otherwise blank.

nathanjmcdougall:
To resolve #519

Preview:
https://pyemu--664.org.readthedocs.build/en/664/


.gitignore
nathanjmcdougall, line 1: If you run `zensical serve`, then you can get  a local hosted docs site which will dynamically re-compile based on changes to the markdown files, to get an interactive experience.

The HTML during the compilation process gets exported to the `site/` directory, which shouldn't be under version control.

docs/Makefile
nathanjmcdougall, line 1: This Makefile was only needed for Sphinx - the zensical CLI can build docs with `zensical build`.

Ditto for `docs/make.bat`

docs/_autosummary/pyemu.ErrVar.rst
nathanjmcdougall, line 1: These docs/**/*.rst` files basically all contained automatically generated API doc skeletons.

The equivalent Zensical config is provided by the `mkdocstrings` dependency in the `api_reference.md` file and is basically just a single line - these skeleton files are not needed.

docs/api_reference.md
```diff
@@ -0,0 +1,3 @@
+# API Reference
+
+::: pyemu
```
nathanjmcdougall: This line uses the `mkdocstrings` package to generate a comprehensive API reference page automatically.

It parses the docstrings, expecting them to be in a consistent format (Google-style for this project).

There are a bunch of warnings/parsing warning emitted by `zensical build` for docstrings it failed to parse. In future efforts, the Ruff ruleset `D4` should help to address a lot of these (with autofixes).

docs/conf.py
nathanjmcdougall, line 1: This `conf.py` file provided configuration for Sphinx. The equivalent is the `zensical.toml` file.

docs/get_pestpp.md
nathanjmcdougall, line 1: I've moved this `get_pestpp.md` file to the docs site - it seems like it is worth including there and it reduces clutter at the root level of the repo.

docs/index.md
nathanjmcdougall, line 1: This `index.md` is the default "landing page" for the site. I've just duplicated some text from the README and provided a few links to other useful material. Over time, this can be improved further, this is just a first pass to get the migration done.

docs/index.rst
nathanjmcdougall, line 1: The `docs/index.md` is basically the new version of this file.

docs/pip.reqs
nathanjmcdougall, line 1: This is replaced by `docs/requirements.txt`. I think this is a more conventional name for the requirements file.

zensical.toml
```diff
@@ -0,0 +1,36 @@
...
+[project.plugins.mkdocstrings.handlers.python]
+inventories = ["https://docs.python.org/3/objects.inv"]
+paths = ["src"]
+
+[project.plugins.mkdocstrings.handlers.python.options]
+docstring_style = "google"
+inherited_members = true
+show_source = false
+
+[[project.theme.palette]]
```
nathanjmcdougall: These theme options basically just set up light-mode/dark-mode handling.

The default is to respect the system theme.

zensical.toml
```diff
@@ -0,0 +1,36 @@
...
+site_url = "https://pyemu.readthedocs.io/"
+nav = [
+  { "Introduction to pyEMU" = "index.md" },
+  { "API Reference" = "api_reference.md" },
+  { "Installing PEST++" = "get_pestpp.md" },
+]
+[project.theme]
+language = "en"
+
+[project.plugins.mkdocstrings.handlers.python]
```
nathanjmcdougall: These `project.plugins.mkdocstrings` sections help automatically generate docs from docstrings

zensical.toml
```diff
@@ -0,0 +1,36 @@
+[project]
+site_name = "pyEMU Documentation"
+site_description = "Documentation for the pyEMU Python package"
+site_url = "https://pyemu.readthedocs.io/"
+nav = [
```
nathanjmcdougall: This `nav` section is where you define references to `.md` files as they should appear in the nav bar on the left-hand side of the site.

Any new `.md` doc files should be declared here.

uv.lock
briochh, line 1: should this be dropped (for now)?

briochh (commented):
Looks like a good option @nathanjmcdougall. Looking at the mockup, it would be good if the API reference could be a bit more broken down through the modules and sub-modules, just so the path through to the commands is a bit clearer. I guess this maybe needs to be defined in the api_reference.md? (Appreciate inconsistencies on the docstring types means that it will not be full populated, initially).

uv.lock
nathanjmcdougall: Yeah that was accidental, I reverted it already I believe.

nathanjmcdougall:
After some experimentation, I think probably Zensical isn't mature enough yet to give what you're looking for in terms of auto-generated API docs.

So I've moved to MkDocs, which is fairly similar but has a more mature plugin ecosystem.

The new preview site is here;

https://pyemu--664.org.readthedocs.build/en/664/

PR #668, opened 2025-12-18
Migrate project management to uv

briochh:
Moving to uv (https://docs.astral.sh/uv/) for project management, build and deployment.

This includes: 
* moving to using pyproject.toml as the main container for all things project config related.
  * project dependancies
  * coverage
 * move back to static versioning but with release initialise in GitHub actions when new version tag is pushed.
 * Removal of a few files that are no longer needed (potentially more can follow)

oh and moving coverage tracking to codecov I think this could be decoupled if that is too offensive.

It's probs not perfect but something to build on?


.coveragerc
briochh, line 1: config moved to pyproject.toml

.github/workflows/ci.yml
```diff
@@ -40,130 +40,88 @@ jobs:
...
+        echo "PATH: $PATH"
+        ls $HOME/.local/bin
+
+    - name: Run tests
+#       For example, using `pytest`
+      # once again need bash (I think) -- windows will run through mingw
       shell: bash -l {0}
       run: |
-        pytest ${{ matrix.test-path }} \
+        uv run pytest ${{ matrix.test-path }} \
```
briochh, line 92: now running tests with uv

.github/workflows/release.yml
briochh, line 1: workflow to publish release -- maybe more motivation here to build some smoke tests

README.md
briochh, line 1: can be refined after merge

_setup.py
briochh, line 1: can be dropped, I think

_versioneer.py
briochh, line 1: can be dropped, no longer using tag based dynamic versioning -- running action on tag push now

autotest/pytest.ini
briochh, line 1: options now in pyproject.toml

pyproject.toml
```diff
@@ -24,72 +21,85 @@ classifiers = [
 ]
 requires-python = ">=3.10"
 dependencies = [
-    "numpy",
-    "pandas",
+    "numpy>2.3.4 ; python_full_version >= '3.14'",
```
briochh: some wrangling of numpy version to try and bypass deadlock issues with certain blas compiles

briochh:
@nathanjmcdougall, you might have some thoughts to add to this PR.

briochh:
@mwtoews, keen to hear your thoughts on this blitz.

pyproject.toml
```diff
@@ -1,17 +1,14 @@
 [build-system]
-requires = [
-    "setuptools >= 61",
-    'tomli; python_version < "3.11"',
-]
-build-backend = "setuptools.build_meta"
+requires = ["uv_build>=0.9.6,<0.10.0"]
```
nathanjmcdougall, line 2: If you're interested in a uv-compatible way of having dynamic versioning for the project, I could set up `hatchling` as the build system instead.

pyproject.toml
```diff
@@ -24,72 +21,85 @@ classifiers = [
...
-versionfile_source = "pyemu/_version.py"
-versionfile_build = "pyemu/_version.py"
-tag_prefix = ""
+[tool.pytest.ini_options]
+testpaths = [
+    "autotest",
+]
+addopts = "--cov-config=pyproject.toml"
 
-[tool.codespell]
```
nathanjmcdougall, line 77: Why remove the codespell config?

pyproject.toml
briochh: I think static is ok. We were using setuptools before but the idea for uv was to bring it under one tool. It seems to be working fine and we can do the dynamic part from GHA when pushing a tag.

.github/workflows/ci.yml
```diff
@@ -40,130 +40,88 @@ jobs:
...
-        scripts/build_lib.sh
-
-    - name: Build pypestutils (non-Windows)
-      if: runner.os != 'Windows'
-      shell: bash -l {0}
-      working-directory: pypestutils
-      run: |
-        scripts/build_lib.sh
+        python-version: ${{ matrix.python-version }}
+        enable-cache: true
```
mwtoews, line 47: By default, cache is enabled for GitHub, so this can be removed. However, one useful option to consider changing is `activate-environment: "true"` to activate a venv ready to be used by later steps. This way you don't need "uv run ..."

_setup.py
mwtoews: This was only used for versioneer. If versioneer is getting the boot, then this can too.

_versioneer.py
mwtoews: can be removed.

.github/workflows/ci.yml
mwtoews: These changes are unnecessary if `activate-environment: "true"` is done as suggested above.

pyproject.toml
briochh: a victim of a shonky manual merge -- should be back in now. Are there dependancies for codespell that should be included? (I never use it...)

pyproject.toml
mwtoews: I'm a personal fan of `["hatchling", "hatch-vcs"]` build backend requirements, since they support dynamic versioning. Using UV does not necessary require using a uv_build build backend.

pyproject.toml
nathanjmcdougall: Yeah, the dependency is just `codespell`. It's quite a useful tool, although probably most useful as a `pre-commit`.

pyproject.toml
```diff
@@ -1,17 +1,14 @@
...
-    'tomli; python_version < "3.11"',
-]
-build-backend = "setuptools.build_meta"
+requires = ["uv_build>=0.9.6,<0.10.0"]
+build-backend = "uv_build"
 
 [project]
 name = "pyemu"
-dynamic = ["version"]
+version = "1.4.0.dev3"
```
briochh: `uv version --bump <BUMP>` is a handy hook for keeping this current, but releases won't be triggered until a git tag is pushed to pypest/pyemu (and the tag starts with 'v')

.github/workflows/release.yml
```diff
@@ -0,0 +1,36 @@
...
+      # TODO: add some smoke tests here
+#      - name: Smoke test (wheel)
+#        run: uv run --isolated --no-project --with dist/*.whl tests/smoke_test.py
+#      - name: Smoke test (source distribution)
+#        run: uv run --isolated --no-project --with dist/*.tar.gz tests/smoke_test.py
+      - name: Publish
+        run: uv publish
+        # The part below with testpypi only
+        # TODO: remove the next two lines for production usage
+        env:
```
briochh: currently just pointing to test.pypi

_versioneer.py
nathanjmcdougall: There's also related versioneer cruft in `pyemu/_version.py` which should be removed

pyproject.toml
nathanjmcdougall: Why `"numpy>=2.2.6,<2.3.2 ; python_full_version < '3.14'",` rather than just `"numpy>=2.2.6 ; python_full_version < '3.14'",`?

It's good to avoid upper bounds. I tried `uv lock` and it resolved fine + tests seem to run. And pre-uv there was no upper bound, and testing was done on Python 3.10.

.github/workflows/release.yml
```diff
@@ -0,0 +1,36 @@
...
+      name: pypi
+    permissions:
+      id-token: write
+      contents: read
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v5
+      - name: Install uv
+        uses: astral-sh/setup-uv@v7
+      - name: Install Python
```
nathanjmcdougall: Why is a python installation step necessary? If using the `uv_build` backend, my understanding is that the build can be done without Python at all.

mwtoews (commented):
I've been using UV for the past year, it's been excellent. A few misc notes below...

.github/workflows/release.yml
```diff
@@ -0,0 +1,36 @@
...
+        run: uv python install 3.13
+      - name: Build
+        run: uv build
+      # TODO: add some smoke tests here
```
nathanjmcdougall, line 27: If you do decide to add smoke tests, I think as a matter of security, you should separate the build+test and publish steps into separate stages. e.g. here

https://github.com/usethis-python/usethis-python/blob/main/.github/workflows/release.yml

The risk is that a compromised test dependency injects a payload into the testing runtime which pre-emptively publishes a malicious package to PyPI, since it's scoped under the `id-token: write` permission. Technically this risk can exist even in the build step, but the only dependency in this case is `uv_build`, so the risk is no higher than using `uv` in the first place.

You can read more about this from the Astral team here

https://docs.pyx.dev/publishing#trusted-publishing

> For isolation reasons, we strongly recommend that you perform your distribution build in a separate job that your publishing job depends on. This ensures that your build environment does not have access to the short-lived credentials used for publishing.

In general (in a follow up PR) it might be good to consider using [zizmor](https://docs.zizmor.sh/) to shore up the GitHub Actions config a bit, from a security perspective

pyproject.toml
briochh: This is a fun one. We have hit a few (possibly different) BLAS deadlock issues on unix runners in numpy.linalg operations (poss related to calls after a fork operation... e.g. https://github.com/numpy/numpy/issues/30092, https://github.com/OpenMathLib/OpenBLAS/issues/5520). 

Pre uv we were pinning mamba install `libopenblas < 0.3.30` to get around a (possibly differentish) issue (but this option not avail on pip installs).

The latest numpy release (2.3.5) points to scipy-openblas 0.3.30.7 which is supposed to fix (https://github.com/numpy/numpy/pull/30157). However, I was still seeing lockups on the CI for py version [3.11, 3.12, 3.13] (and py 3.10 not supported by numpy 2.3.5) so think that we were still suffering from the previous issue that led us to pin < 0.3.30.

So in summary, we seem to need openblas pre 0.3.30 for py < 3.14 but py 3.14 works with a fixed openblas formulation (0.3.30.7) which is only picked up by numpy 2.3.5 (I think py 3.14 only works for us with this option). I have been around the houses a few times on this (without fully understanding it) and this seemed to be the dependancy combo that works. I didn't want to go down the route of local compile of BLAS in the CI.

Note: only an issue on Unix runners Windows and macOS use different math libs.

nathanjmcdougall: Thanks for explaining, that all makes sense.

Since it seems to be OS-specific, perhaps you could add further specifiers to restrict the case handling to just linux:

```
"numpy>2.3.4 ; python_full_version >= '3.14' and sys_platform=='linux'"
"numpy>=2.2.6,<2.3.2 ; python_full_version < '3.14' and sys_platform=='linux'",
"numpy>=2.2.6 ; sys_platform!='linux'"
```

.github/workflows/ci.yml
briochh: my understanding is that using `uv run` explicitly is preferable (https://github.com/astral-sh/setup-uv/blob/main/docs/environment-and-tools.md#activate-environment) unless there is some other benefit that I am unaware of?

mwtoews: Using `uv run` is fine, keep the default of not active the environment.

pyproject.toml
briochh: This is bit of blindspot for me. I was enjoying the idea of stripping out the need to another tool in the project. And that uv build backend is rapid as. Currently we are using setuptools with versioneer to deal with all the dynamic versioning complexity. Moving to uv and the uv_build backend seemed to simplify things a lot but with the cost of moving version to static metadata. I understand that dynamic metadata is a slightly "unpreferred" option in some quarters. The GHA action seems like a good way to tie versioning back to the VCS. I am open to suggestions and improvements to this but most keen on least worst, simplest option!

_versioneer.py
briochh: sufficient to just point to 
`__version__ = importlib.metadata.version(__name__)`
in `__init__.py`?

pyproject.toml
briochh: will try that out

pyproject.toml
nathanjmcdougall: You will always need _a_ build backend, it's just a matter of choosing which one. `uv_build` is treated specially by `uv` because it doesn't need to spin up Python processes to run the build, it can just run from within `uv`. That makes it faster than hatchling but in my experience, both are fast enough that the difference is negligible.

Using `hatchling` with `hatch-vcs` allows for the same sort of dynamic versioning as `versioneer` although it's a more modern configuration option. Hatchling used to be the default build backend in `uv` before `uv_build` was an option.

This is what hatchling with `hatch-vcs` config looks like:

```TOML
[build-system]
build-backend = "hatchling.build"
requires = [
  "hatch-vcs>=0.5.0,<0.6.0",
  "hatchling>=1.27.0,<1.28.0",
]

[project]
dynamic = [ "version"]

[tool.hatch.version]
source = "vcs"

[tool.hatch.build.hooks.vcs]
version-file = "pyemu/_version.py"
```

Versus with `uv_build`:

```TOML
[build-system]
build-backend = "uv_build"
requires = ["uv_build>=0.9.6,<0.10.0"]

[project]
version = "1.4.0.dev3"
```

So they're very similar but you don't need to maintain the `.dev3` suffixes manually etc. if you want to do local builds. Personally I'd say that would be the main driver for whether you'd benefit from dynamic versioning - if you like to do a lot of local wheel builds and distribute dev versions, to disambiguate different wheel files.

With `hatch-vcs`, if you build then it will use git tags to determine the version of the wheel, adding `dev` suffixes etc. based on the commit hash. It will also create the `__version__` value in `pyemu/_version.py` (or wherever you specify)

briochh: leaving with uv_build back end for now -- I think it works for our purposes happy for improvements to be made at a later date though...

.github/workflows/release.yml
briochh: on "the list"

PR #672, opened 2026-01-12
Feat lesswarn

jtwhite79:
less warning, etc.  trying to improve the notebook experience...


pyemu/pst/pst_handler.py
briochh, line 1: guess we could do better with a logger in the pst_handler but for now I think this is fine

pyemu/utils/helpers.py
briochh, line 1: dito -- maybe a logger option in future (if anyone cares)

briochh (approved):

PR #675, opened 2026-01-23
Feat applyrefac

briochh:
Refactoring apply methods, starting with pilot point process orienting to PPU. 
Trying to separate python-based and ppu based calls for easier decoupling later.

Also dropping top level emulator class imports as I think it is slowing down pyemu import, potentially a lot when multiprocessing...


briochh:
Potential breaking changes:
 - If pypestutils is installed and the `PestUtilsLib` is importable then `OrdinaryKrige` will alias to `OrdinaryKrigePPU`. This means that previous, pyemu and python-based methods are not accessible from `OrdinaryKrige` class object (e.g. `calc_factors_grid`). To force use of old methods, class `OrdinaryKrigeOrg` needs to be explicitly called (unless ppu is not available).
- PPU based factor calcs and factor apply (fac2real equivalent) now default to a binary factors file [it is much faster to write and read] -- an optional argument `fac_ftype` can be passed to the calc_factors method to write ascii.
- Multiple variograms in geostruct are no longer supported (when using ppu methods)
- Emulator classes are no longer available from package top level (e.g. `pyemu.DSI`), should be accessed through pyemu.emulators.DSI instead
- Probs more

briochh:
- [x]  Need to check method of propagating fac_ftype to fac2real methods

pyemu/utils/pst_from.py
```diff
@@ -2914,12 +2968,16 @@ def add_parameters(
...
+    add_pars = add_parameters
+
+    # def _check_pp_fac_calc(self):
 
     def _prep_pp_args(self, zone_array, pp_kwargs=None):
         if pp_kwargs is None:
             pp_kwargs = dict([])
 
         if not pp_kwargs["use_pp_zones"]:
+            zone_array = zone_array.copy()  # copy to preserve original
```
jtwhite79, line 2982: Does this preserve the org?  Seems like they have the same name?

jtwhite79 (commented):
All looks good to me - just one question about copying the zone_array.

pyemu/utils/pst_from.py
briochh: I think this preserves the original. I was having issues with the array changing in the parent function so threw this in. Maybe it is lazy. Is there a better way?

PR #676, opened 2026-01-27
Support extra imports like: from IMPORT import MODULE

p-ortega:
@briochh Quick fix with an if statement to be able to add imports to the forward_run.py like:

`from IMPORT import MODULE`

Open to ideas

### DIFF
--- a/pyemu/utils/pst_from.py
+++ b/pyemu/utils/pst_from.py
@@ -543,8 +543,10 @@ class PstFrom(object):
             )
             f.write("import pyemu\n")
             for ex_imp in self.extra_py_imports:
-                f.write("import {0}\n".format(ex_imp))
-
+                if ex_imp.startswith("from"):
+                    f.write("{0}\n".format(ex_imp))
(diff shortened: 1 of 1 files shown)


p-ortega:
As draft for now waiting for @briochh opinion

pyemu/utils/pst_from.py
```diff
@@ -543,8 +543,10 @@ def write_forward_run(self):
             )
             f.write("import pyemu\n")
             for ex_imp in self.extra_py_imports:
-                f.write("import {0}\n".format(ex_imp))
-
+                if ex_imp.startswith("from"):
```
briochh: could we be even more flexible with `ex_imp.startswith("from ") or ex_imp.startswith("import ")`? (note the space after the string, just in case for packages like importlib!)

p-ortega: sounds good, will cook something

briochh:
thanks @p-ortega

PR #678, opened 2026-02-03
feat emulator_generalization

rhugman:
- refactored Base emulator class to make pestpp setup functions more generalizable
 - more tests
 - updated GPR to enable runstor runs


mnfienen:
hey @rhugman - I ran with this building GPR using the runstor option. #nice! But.....it seems like prior information from my initial PST file gets dropped but I'm using for objectives. I think in the develop branch that wasn't the case?

rhugman:
> hey @rhugman - I ran with this building GPR using the runstor option. #nice! But.....it seems like prior information from my initial PST file gets dropped but I'm using for objectives. I think in the develop branch that wasn't the case?

@mnfienen - lemme know if this resolves it: https://github.com/pypest/pyemu/pull/678/commits/79ff227ae01dd5659e493bdcd357b53e01408e68

briochh:
@mnfienen looks like this should work for prior information now? Can you confirm? @rhugman how easy would it be to make sure that is covered in a test?

mnfienen:
hey @briochh  and @rhugman  --- sorry for the delay. Just confirmed that fix worked on the PI - thanks!

I think an easy test would be to run `prepare_pestpp` on an original PST file that contains PI and then assert that the GPR PST file has a PI external csv file after writing out the PST file version==2 style.

I think a common PI case is what I have - just the sum of all the decvars (pumping rates). So like:
``` 
newpst.add_pi_equation(decvars, # parameter names to include in the equation
                    pilbl="obj_well",  # the prior information equation name
                    obs_group="less_than_pumping",
                    rhs=1e13) # note the "less_" prefix.
```

mnfienen:
actually, I was able to cook up a test and just assert that `pst_gen.prior_information is not None` after running `prepare_pestpp` on a PST object that has prior information equations added to it. Test is included in my commit.

briochh:
Thanks @rhugman and @mnfienen, I'll get this merged. May be good to look at dialling in some more tests, although I guess efficiency might be hard...? Most of the added froward_run helpers are not being picked up by the coverage maybe this is because the are run by and external instance, or we are not testing?

PR #680, opened 2026-02-14
Fix shapefile field type handling for pandas StringDtype

spark-hydro:
## Description
Fixes issue with shapefile Writer when pandas DataFrames use `StringDtype` instead of object dtype for string columns.

## Problem
When using pandas 3.0+ string columns (dtype='string' or `pd.StringDtype()`), the shapefile writer raised:
```
unrecognized field type in par_info:name:str
```

This occurs because pandas 3.0 introduced `StringDtype` as the default for string columns, which returns `'str'` when converted to string, rather than `'object'` as in previous pandas versions.

## Solution
Added check for `str(dtype) == 'str'` to properly handle pandas `StringDtype` columns alongside the existing object dtype check.

## Changes
- Modified dtype checking in `pp_utils.py` to recognize `str(dtype) == 'str'`
- Maintains backward compatibility with older pandas versions using object dtype

### DIFF
--- a/pyemu/utils/pp_utils.py
+++ b/pyemu/utils/pp_utils.py
@@ -289,7 +289,9 @@ def setup_pilotpoints_grid(
         except:
             shp = shapefile.Writer(shapeType=shapefile.POINT)
         for name, dtype in par_info.dtypes.items():
-            if dtype == object:
+            dtype_str = str(dtype)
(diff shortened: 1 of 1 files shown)


briochh:
Thanks @spark-hydro, there is a good chance there may be more of these in the codebase

PR #684, opened 2026-02-19
Fix smp_to_ins for SMP files with leading whitespace

### LINKED ISSUE #361: Issue with method "smp_utils.smp_to_ins" in pyEMU
Is there an issue with the "smp_utils.smp_to_ins" method in pyEMU? I.e., due to a leading whitespace added in front of the observation name as the first character in each line of the .smp file created by mod2smp, the generated instruction files look erroneous. I.e., (1) if using the free format then one "w" marker is missing (only 3 instead of 4), and (2) if the fixed format is used, then the data field is shifted one column leftward (i.e., to the time column instead of the actual data (e.g., head) column).

reneangermeyer:
## Summary

Fixes the `smp_to_ins` instruction file generation for SMP files with leading whitespace (e.g. output from `mod2smp`).

The PEST `w` instruction uses a two-step cursor algorithm where leading whitespace consumes one `w` marker without advancing past a word. This caused `smp_to_ins` to generate incorrect instruction files:

- **Free format** (`gwutils_compliant=False`): now uses 4 `w` markers when leading whitespace is detected (3 when not), so the value column is correctly read instead of the time column.
- **Fixed format** (`gwutils_compliant=True`): dynamically determines the value column range from the actual SMP file layout instead of hardcoding `39:46`, which only worked for short site names without leading whitespace.

## Changes

- `pyemu/utils/smp_utils.py`: reads raw SMP lines to detect leading whitespace and compute correct column positions
- `autotest/utils_tests.py`: adds `smp_to_ins_leading_whitespace_test` covering both modes with and without leading whitespace, plus the existing repo SMP test files

## Test plan

- [x] New test `smp_to_ins_leading_whitespace_test` passes
- [x] Existing tests `smp_test`, `smp_to_ins_test`, `smp_dateparser_test` still pass

Fixes #361

### DIFF
--- a/autotest/utils_tests.py
+++ b/autotest/utils_tests.py
@@ -2019,6 +2019,120 @@ def smp_test(tmp_path):
     print(len(obs_names))
 
 
+def smp_to_ins_leading_whitespace_test(tmp_path):
+    """Test that smp_to_ins handles leading whitespace correctly in both
+    free format (gwutils_compliant=False) and fixed format
+    (gwutils_compliant=True) modes.
+
+    The PEST 'w' instruction treats leading whitespace differently:
+    it consumes one 'w' to skip past leading blanks without advancing
+    past a word. Files with leading whitespace need an extra 'w' marker
+    in free format, and correct column ranges in fixed format.
+
+    See: https://github.com/pypest/pyemu/issues/361
+    """
+    import os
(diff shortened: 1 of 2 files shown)


briochh:
Thanks for this @reneangermeyer

PR #698, opened 2026-05-21
fix in results handler for the insanity of mixed-case "case" names

jtwhite79:


pyemu/utils/helpers.py
```diff
@@ -2040,6 +2040,8 @@ def _process_array_file(model_file, df):
                 continue
             if str(mlt).endswith(".npy"):
                 mlt_data = np.load(mlt)
+                if mlt_data.ndim == 1:
```
jtwhite79, line 2043: @briochh you ok with this change?  when model is unstructured and using ppu for pilotpoints, the arr from the npy was coming back 1-D...

briochh: Thought that catch was already in there. Makes sense.

